PR #12, opened 2024-09-18
Add comment to scan()

v-jkegler:
Begin the comment for scan()

### DIFF
--- a/parser/src/earley/parser.rs
+++ b/parser/src/earley/parser.rs
@@ -1100,7 +1100,17 @@ impl ParserState {
         }
         true
     }
+    
+    // scan() implements the version of Earley described in the
+    // slides for "Parsing: Earley parsing", Winter 2017/2018,
+    // Laura Kallmeyer, Heinrich Heine Universitaet, Dusseldorf,
+    // https://user.phil-fak.uni-duesseldorf.de/~kallmeyer/Parsing/earley.pdf
+    // (Retrieved 18 Sep 2024).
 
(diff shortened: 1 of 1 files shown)


mmoskal (approved):
LGTM

parser/src/earley/parser.rs
```diff
@@ -1100,7 +1100,17 @@ impl ParserState {
...
+    
+    // scan() implements the version of Earley described in the
+    // slides for "Parsing: Earley parsing", Winter 2017/2018,
+    // Laura Kallmeyer, Heinrich Heine Universitaet, Dusseldorf,
+    // https://user.phil-fak.uni-duesseldorf.de/~kallmeyer/Parsing/earley.pdf
+    // (Retrieved 18 Sep 2024).
+
+    // An important difference between the algorithm implemented here
+    // and Kallmeyer's is that in scan(), the token scan is performed
+    // first, while in Kallmeyer it is performed last.
```
hudson-ai, line 1112: Does this difference lead to any changes in resulting behavior (or otherwise any real "practical" difference)?

mmoskal: I think it's only presentation, ie in which iteration of the loop you consider the steps

v-jkegler: Actually, it's an important difference.  The order implemented by scan() enables [the Ruby Slippers technique](https://blogs.perl.org/users/jeffrey_kegler/2010/06/parsing-with-ruby-slippers.html).

PR #18, opened 2024-09-26
Add comments to parser.rs

v-jkegler:
Add comments to parser.rs

### DIFF
--- a/parser/src/earley/parser.rs
+++ b/parser/src/earley/parser.rs
@@ -246,11 +246,14 @@ impl Scratch {
         }
     }
 
+    // Set current working Earley set to the single item
+    // at 'pos' in the Earley table.
     fn new_row(&mut self, pos: usize) {
         self.row_start = pos;
         self.row_end = pos;
     }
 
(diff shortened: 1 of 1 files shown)


parser/src/earley/parser.rs
```diff
@@ -246,11 +246,14 @@ impl Scratch {
         }
     }
 
+    // Set current working Earley set to the single item
```
mmoskal, line 249:
```suggestion
    // Set current working Earley to empty set
```

parser/src/earley/parser.rs
```diff
@@ -246,11 +246,14 @@ impl Scratch {
...
+    // at 'pos' in the Earley table.
```
mmoskal, line 250:
```suggestion
    // The set backing data is at `pos`
```

mmoskal:
I did a small fix - new_row() is empty

v-jkegler:
Good you caught this.  And it makes sense that the code should work that way.  Thanks!
________________________________

PR #19, opened 2024-09-26
row_start, row_end comment in parser.rs

v-jkegler:
row_start, row_end comment in parser.rs


v-jkegler:
This comment simply states a Rust convention, which anyone reading or writing Rust code should be expected to know.  It's clutter and I am withdrawing it.

v-jkegler:
Closing for the reason given in the previous comment.

PR #30, opened 2024-10-28
Add args to enable/disable backtracking and fast-forwarding in LLInterpreter

JC1DA:
Note: This is helpful to integrate llguidance into inference frameworks that do not support backtracking atm.


mmoskal:
Looks good, but please remove the conditional_ff_tokens - they are not used and will just confuse people!

lochuynh1412:
@mmoskal should I leave the default value to true or set it to value of enable_ff_tokens?

JC1DA:
> @microsoft-github-policy-service agree

@microsoft-github-policy-service agree

Issue #44, opened 2024-11-08
port json_schema -> llguidance from Guidance project


mmoskal:
Currently Guidance has the best json schema coverage of all constrained encoding libraries. However, the code translating from json schema to llguidance format sits in Python Guidance and can't be used natively.

While basic json schema features are supported, the more advanced are yet to be ported.

Filing this issue for tracking and linking to it.

PR #48, opened 2024-11-09
Port guidance's JSON implementation

hudson-ai:
Working towards resolving #44 

Super draft PR, just opening for feedback and help with my extremely amateurish rust (thanks @mmoskal!)

To run guidance's JSON tests with this as the backend, check out my (also very rough) guidance branch
https://github.com/hudson-ai/guidance/tree/llguidance_json


hudson-ai:
@microsoft-github-policy-service agree company="Microsoft"

parser/src/json/numeric.rs
```diff
@@ -0,0 +1,364 @@
...
+fn float_to_str(f: f64) -> String {
+    format!("{}", f)
+}
+
+pub fn rx_float_range(
+    left: Option<f64>,
+    right: Option<f64>,
+    left_inclusive: bool,
+    right_inclusive: bool,
+) -> String {
```
mmoskal, line 222: Should be easy enough to use `anyhow::Result<String>` and than `ensure!()` instead of `assert!()` and `bail!()` instead of `panic!()`, at least in the top-level function. The inner asserts should probably never happen.

mmoskal (approved):
overall looks great!

parser/src/json/compiler.rs
```diff
@@ -118,6 +117,13 @@ macro_rules! cache {
 }
 
 impl JsonCompileOptions {
+    pub fn default() -> Self {
```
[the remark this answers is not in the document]
hudson-ai: Got it, thanks! Didn't realize it was tied to a trait nor that it was already being derived

hudson-ai:
> overall looks great!

Thanks! Please don't approve until it's done though 😆 
That being said, I can totally split this into several PRs.

parser/src/json/compiler.rs
```diff
@@ -614,52 +614,57 @@ impl Compiler {
             items.push((seq, false));
         }
         let opener = self.builder.string("{");
-        let inner = self.ordered_sequence(&items, false);
+        let inner = self.ordered_sequence(&items, false, &mut HashMap::<(&[(NodeRef, bool)], bool), NodeRef>::new());
```
mmoskal, line 690: &mut HashMap::new() should work without types?

parser/src/json/compiler.rs
```diff
@@ -607,42 +609,60 @@ impl Compiler {
             let is_required = required.contains(name);
             // Quote (and escape) the name. TODO: probably overkill to use json_dumps here
             let quoted_name = json_dumps(&json!(name));
-            if property_schema == &Value::Bool(false) {
-                if is_required {
-                    bail!("Required property has 'false' schema: {}", name);
+            let property = match self.gen_json(property_schema) {
+                Ok(node) => node,
+                Err(e) => match e.downcast_ref::<UnsatisfiableSchemaError>() {
+                    // If it's not an UnsatisfiableSchemaError, just propagate it normally
```
mmoskal, line 637: funky!

parser/src/json/compiler.rs
hudson-ai: Was having trouble with an earlier state of this code -- hopefully it can do the type inference now 😆

parser/src/json/compiler.rs
hudson-ai: Was just about to comment "funky good or funky bad?" but I suppose you answered that 😆

parser/src/json/schema.rs
```diff
@@ -247,6 +251,7 @@ struct Context<'a> {
     // TODO: actually use this to handle draft-specific behavior
     draft: Draft,
     defs: Rc<RefCell<HashMap<String, Schema>>>,
+    seen: Rc<RefCell<HashSet<String>>>,
```
mmoskal, line 254: at some point, you will probably want `Rc<RefCell<InnerContext>>` (probably when adding normalize counters)

hudson-ai: Oh thanks, I'll look into it. Was just about to start adding said counters to track the number of Schema nodes that are being constructed. Probably not to normalize itself though, as I just made that "shallow".

hudson-ai:
@mmoskal re: consts and enums...

"Upstairs" in the grammar compiler, I'm already transforming non-atomic const nodes (objects and arrays) into strict schemas that are then being passed to gen_json_*. This is so that they're actually properly lexed and have the correct whitespace behavior.

I could do this earlier in the Schema builder, doing it for "atomic" types too, e.g.
```
{"const": 42} -> {"type": "integer", "minimum": 42, "maximum": 42}
```
This would allow our existing intersection logic to handle these cases rather than using a full JSON validator. It would be more limited in its capabilities, but it would solve the problem of not having a real schema object when we're merging.

The only complication here is bools, which have no built-in mechanism for constraining them. So I think I would just have to add a LiteralBool type.

hudson-ai:
Hoping to merge this ASAP though, and I think I will worry about consts, enums, and refs in a follow-up PR. Will work on counters and unnecessary cloning today

mmoskal:
I like the const->schema idea. We just need to make sure we get similarly efficient schemas from out min/max etc logic. 

Also, for the case of enum it would be ideal to generate a single regex for the enum, not a grammar (grammars have a limit of at most ~200 items in a select, but regexes don't).

But all good for later PRs, let's marge this!

PR #60, opened 2024-11-21
fix the pyi to reflect the correct signature of JsonCompiler

hudson-ai:
The signature of JsonCompiler in the .pyi was wrong.

@lochuynh1412 fyi

### DIFF
--- a/python/llguidance/_lib.pyi
+++ b/python/llguidance/_lib.pyi
@@ -143,7 +143,7 @@ class LLInterpreter:
 class JsonCompiler:
     def __new__(
         cls,
-        separators: Optional[tuple[str, str]] = None,
+        separators: Optional[Tuple[str, str]] = None,
         whitespace_flexible: bool = False,
     ) -> "JsonCompiler":
         """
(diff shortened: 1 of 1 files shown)


mmoskal (approved):
also, in py.rs

```
    #[pyo3(signature = (separators = None, whitespace_flexible = false))]
    fn py_new(separators: Option<(String, String)>, whitespace_flexible: bool) -> Self {
```

i think the following would have the same effect:

```
    fn py_new(separators: Option<(String, String)>, whitespace_flexible: Option<bool>) -> Self {
```

hudson-ai:
> I think the following would have the same effect:
> 
> ```
>     fn py_new(separators: Option<(String, String)>, whitespace_flexible: Option<bool>) -> Self {
> ```

Interesting. Are you sure that's the same? From the python-side, I don't want to accept None as a value for whitespace_flexible, it's a boolean with a default value. Need to read a bit of py03 documentation to see how this stuff *actually* works.

mmoskal:
you're probably right; I sort of assumed passing `None` and not passing arg should behave the same from the user PoV, but maybe that's not how you do things in Python-land.

One could clearly see the advantage of having both null and undefined in JS! Well... not really...

PR #61, opened 2024-11-21
Add tests for numeric regular expressions; some bug fixes

hudson-ai:
Ports the numeric range tests we're currently doing in python into the crate. Found some bugs and fixed them.


hudson-ai:
> If needed, there is `rstest` crate which lets you do fixtures and cases. Never used it yet though.

Cool, may very well look into it!

PR #62, opened 2024-11-22
Add lark and regex compilers to the python API

hudson-ai:
Adds `LarkCompiler` and `RegexCompiler` to the python API. Also changes the `JsonCompiler` slightly such that it returns a `TopLevelGrammar` rather than a `GrammarWithLexer` node, such that the output is "ready to consume" by an `LLInterpreter`.

@lochuynh1412 should make your life easier!


hudson-ai:
@mmoskal currently both the Lark and the Regex compilers take no args on instantiation, so they feel pretty pointless at the moment. But I felt that having an API that is consistent with JSON (which *does* take args) was nice.

Open to suggestions for how to make this API nicer so that we don't get stuck with something ugly.

mmoskal (approved):
I think it's OK to have the classes

PR #63, opened 2024-11-22
align API names with Constraint

mmoskal:


hudson-ai (approved):
LGTM

hudson-ai:
@mmoskal is `commit_token` returning None on the last call the only real semantic change here beyond renaming the methods?

hudson-ai:
@lochuynh1412 pinging you for visibility

mmoskal:
I reverted the `commit_token()` change, since it was returning stop only in case `compute_mask()` already did (in which case we don't even call it); easier to keep it as it was

Issue #58, opened 2024-11-21
Grammar rules should be unique


v-jkegler:
Currently there is no mechanism in llguidance for determining a grammar rule (aka node) is unique.  This

1. makes the algorithm for counted sequences sub-optimal, and may prevent it from reducing the number of rules to O(n) in the length of the sequence;
2. makes our definition of grammar unnecessarily inconsistent with the usual one on the world of parsing theory, which
3. is likely, as we add parsing features, to cause other gotchas.

mmoskal:
This can be done in `GrammarBuilder`; I guess we should also try to do it in `Grammar::optimize`

hudson-ai:
@mmoskal linked issue for hash consing?

mmoskal:
I think this would be for both hash-consing in the GrammarBuilder and also Hash of (node_id, num_repetitions) cache in repeat_exact and (a separate but same shape) in at_most

v-jkegler:
See Commit 5bc30a8

hudson-ai:
@v-jkegler uniqueness is enforced a bit further up the stack when we are building a "TopLevelGrammar", see my recent commit here
https://github.com/microsoft/llguidance/blob/29655f3aef74a21fcdbfa7c4e349c07baa89e62d/parser/src/grammar_builder.rs#L186

PR #70, opened 2024-11-25
Cache `NodeRef`s in `grammar_builder`

hudson-ai:
- For the FHIR schema, all the serialization calls slow things down by about 15%, (from ~140ms to build the whole grammar to ~160ms)
- For the FHIR schema, this decreases the size of the compiled grammar by about 30%
- I have concerns about storing serialized grammars as keys, as this should about double the size of our `GrammarBuilder` struct. But this is pretty inconsequential unless we're talking huge schemas like FHIR (and I suppose 2*.7=1.4 means we're not quite doubling the size overall)

There are likely quite a few optimizations we can still do to reduce the time to build grammars, both in the JSON compiler and in `repeat`.

Linked to issue #58

### DIFF
--- a/parser/src/grammar_builder.rs
+++ b/parser/src/grammar_builder.rs
@@ -177,7 +177,7 @@ impl GrammarBuilder {
 
     pub fn add_node(&mut self, node: Node) -> NodeRef {
         // Generate a key for the node from its serialized form if it is not the placeholder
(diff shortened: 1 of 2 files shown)


hudson-ai:
@mmoskal ready for review :)

mmoskal (approved):

PR #72, opened 2024-11-26
Change message about calling compute_bias() twice

v-jkegler:
Change message about calling compute_bias() twice

### DIFF
--- a/parser/src/constraint.rs
+++ b/parser/src/constraint.rs
@@ -158,7 +158,7 @@ impl Constraint {
         ensure!(!self.last_res.is_stop(), "compute_bias() called after stop");
         ensure!(
             self.step_arg.is_some(),
-            "commit_token() not called before compute_bias()"
+            "compute_bias() called twice without intervening commit_token()"
         );
         let step_arg = self.step_arg.take().unwrap();
         let res = self.parser.mid_process(step_arg);


parser/src/constraint.rs
```diff
@@ -158,7 +158,7 @@ impl Constraint {
         ensure!(!self.last_res.is_stop(), "compute_bias() called after stop");
         ensure!(
             self.step_arg.is_some(),
-            "commit_token() not called before compute_bias()"
+            "compute_bias() called twice without intervening commit_token()"
```
hudson-ai, line 161: Can you change this error message to reflect the (I assume new) function name "compute_mask"? There are a few occurrences of the name "compute_bias" before this too.

v-jkegler: Both a compute_mask() and a compute_bias() are present in the code, one calling the other.  You and @mmoskal should do whatever you think appropriate here.

v-jkegler: @hudson-ai If you haven't already, you might want to consult the chat for yesterday, where @mmoskal and I had a brief discussion of this message.  It was that discussion which motivated me to file this issue.

mmoskal:
I have a bunch of other changes pending around this and will work more on this today. Will get back to this PR when I'm done!

parser/src/constraint.rs
v-jkegler: @hudson-ai No problem.  This is a very easy call stack to get lost in!

mmoskal:
the change is already applied, thanks!

PR #73, opened 2024-11-26
Add commentary on the mid-process methods

v-jkegler:
Add commentary on the mid-process methods.


mmoskal:
Please see the new APIs in tokenparser.rs - now consuming tokens and computing mask are decoupled

PR #78, opened 2024-12-02
Update json README.md

hudson-ai:


mmoskal (approved):
we probably should "stabilize" the allOf order at some point

PR #83, opened 2024-12-08
make timezone optional

mmoskal:
ChatGPT told me: In JSON Schema, the date-time format does not explicitly require a timezone, but it is recommended. According to the [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339) standard, which JSON Schema date-time format adheres to, a timezone can be either specified (e.g., Z for UTC or a specific offset like +02:00) or omitted.

both python jsonschema and https://www.jsonschemavalidator.net/ let one omit it but Rust jsonschema requires it

@hudson-ai any opinion?


mmoskal:
OK, never mind. Just checked the RFC. Our incoming AI overlord is wrong, so is Python and that website. @hudson-ai and Rust library are right.

PR #88, opened 2024-12-10
Treat UTF-16 surrogate pairs as single characters for string min/maxLength

hudson-ai:
Makes the following test pass:
```json
  {
      "description": "minLength validation",
      "schema": {
          "$schema": "https://json-schema.org/draft/2020-12/schema",
          "minLength": 2
      },
      "tests": [
          {
              "description": "one grapheme is not long enough",
              "data": "\uD83D\uDCA9",
              "valid": false
          }
      ]
  }
```

### DIFF
--- a/parser/src/json/compiler.rs
+++ b/parser/src/json/compiler.rs
@@ -173,7 +173,7 @@ impl Compiler {
                 message: reason.to_string(),
             })),
             Schema::Null => Ok(self.builder.string("null")),
-            Schema::Boolean => Ok(self.lexeme(r"true|false")),
+            Schema::Boolean => Ok(self.lexeme(r"true|false", false)),
             Schema::Number {
                 minimum,
                 maximum,
@@ -285,11 +285,11 @@ impl Compiler {
         }
     }
 
-    fn lexeme(&mut self, rx: &str) -> NodeRef {
+    fn lexeme(&mut self, rx: &str, json_quoted: bool) -> NodeRef {
         if self.lexeme_cache.contains_key(rx) {
             return self.lexeme_cache[rx];
         }
-        let r = self.builder.lexeme(mk_regex(rx), false);
+        let r = self.builder.lexeme(mk_regex(rx), json_quoted);
         self.lexeme_cache.insert(rx.to_string(), r);
         r
     }
@@ -333,7 +333,7 @@ impl Compiler {
                 minimum, maximum
             )
         })?;
-        Ok(self.lexeme(&rx))
+        Ok(self.lexeme(&rx, false))
     }
 
     fn json_number(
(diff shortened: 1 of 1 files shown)


parser/src/json/compiler.rs
```diff
@@ -352,11 +353,11 @@ impl Compiler {
                     minimum, maximum
                 )
             })?;
-        Ok(self.lexeme(&rx))
+        Ok(self.lexeme(&rx, false))
     }
 
     fn json_simple_string(&mut self) -> NodeRef {
-        self.lexeme(&format!("\"{}*\"", CHAR_REGEX))
+        self.lexeme("(?s:.*)", true)
```
hudson-ai, line 360: Note that this change is orthogonal to the underlying PR -- @mmoskal if using the `CHAR_REGEX` directly is marginally more performant, I can switch it back.

parser/src/json/compiler.rs
```diff
@@ -568,12 +569,14 @@ impl Compiler {
...
         } else {
-            Ok(self.lexeme(&format!(
-                "\"{}{{{},{}}}\"",
-                CHAR_REGEX,
-                min_length,
-                max_length.map_or("".to_string(), |v| v.to_string())
-            )))
+            Ok(self.lexeme(
+                &format!(
+                    "(?s:.{{{},{}}})",
```
hudson-ai, line 574: `derivre` seems smart enough to match `\uD83D\uDCA9` with `.`, so length is counted appropriately

mmoskal:
The JSON-quoted derivre strings do not allow `\uxxxx` escapes except for `\x00xx`. The assumption is that everything else will be emitted as UTF-8.

It's all the same as far as JSON goes (when you read JSON `{"x":"ł"}` or `{"x":"\u0142"}` you get exactly the same object in memory), but I don't know if we want to let the model output the general `\uxxxx` sequences in non-regex case. I would say not just to be consistent. If we do allow it, we would also need to check if the surrogate pairs with `\xD8xx` are output correctly (that is the high surrogate only ever occurs before low surrogate and vice versa).

mmoskal:
Oh and BTW the test will pass if do `json.dumps(..., ensure_ascii=False)`

hudson-ai:
@mmoskal thanks for taking a close look at this. I have to admit that all of this encoding stuff is still very far from intuitive to me...

RE: being self-consistent... I would, as a user, expect
```
{"maxLength": X, "minLength": Y}
```
to produce an identical constraint to
```
{"pattern": "(?s:.{X,Y})"}
```

Does this imply that we should use my solution but change the JSON quoting semantics in derivre? Or something else?

mmoskal:
A JSON object `{"x":"ł"}` can be also encoded as `{"x":"\u0142"}`. It is the same JSON object when read into memory, just a different serialization format (in Python it's the `ensure_ascii` option, which defaults to `False` (`\u0142`); most other libraries/languages AFAIK default to `ł` as it is shorter and easier to read). In some sense this is similar to `{ "x" : "\u0142" }` (with spaces) also being the same object.

Now, the model may want to generate `ł` or `\u0142`; I don't know which one it likes more, but my money would be on `ł`, Who knows if the test data was passed through Python with default settings at some point, but I suspect not.

Our current unconstrained string regex allows for both, while derivre JSON-quoted only allows for `ł`. I would say let's merge this and if we find that models like `\u0142`, we'll change derivre.

mmoskal (approved):
maybe rename to "disallow surrogate pairs in strings with min/maxLength"

PR #91, opened 2024-12-11
optimize enums

mmoskal:
Not tested at all, but that was my idea

### DIFF
--- a/parser/src/json/schema.rs
+++ b/parser/src/json/schema.rs
@@ -10,6 +10,8 @@ use referencing::{Draft, Registry, Resolver, ResourceRef};
 use regex_syntax::escape;
 use serde_json::Value;
 
+use crate::api::RegexNode;
+
 const DEFAULT_ROOT_URI: &str = "json-schema:///";
 const DEFAULT_DRAFT: Draft = Draft::Draft202012;
 const TYPES: [&str; 6] = ["null", "boolean", "number", "string", "array", "object"];
@@ -95,6 +97,7 @@ pub enum Schema {
         max_length: Option<u64>,
         pattern: Option<String>,
         format: Option<String>,
+        const_string: Option<String>,
     },
     Array {
         min_items: u64,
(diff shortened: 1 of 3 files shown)


hudson-ai:
Overall looks pretty good! I think there's a few edge cases we need to watch out for though, particularly when intersecting with other constraints. E.g. intersecting a constant string with a format or pattern may yield an unsatisfiable (sub)schema

parser/src/json/schema.rs
```diff
@@ -128,6 +131,40 @@ impl Schema {
...
+            Schema::Boolean => return Some(RegexNode::Regex("true|false".to_string())),
+            Schema::Number {
+                minimum: Some(x),
+                maximum: Some(y),
+                ..
+            } if x == y => return Some(RegexNode::Literal(x.to_string())),
+            Schema::String {
+                const_string: Some(s),
+                ..
+            } => s,
```
hudson-ai, line 146: As long as you ensure that there are no other constraints (e.g. due to intersection with siblings) that might invalidate this const_str. Could maybe handle that in the intersection logic, but it should be sufficient to change this to

```rust
            Schema::String {
                min_length: 0,
                max_length: None,
                const_string: Some(s),
                pattern: None,
                format: None,
            } => s,
```

hudson-ai: Passes guidance tests (which include some weird cases where other constraints are combined with enums) with this change

hudson-ai: but of course this is totally mitigating the entire point of the const, as we're adding a regex when we add the const string... needs modification

mmoskal: my idea was that `const_string` means it's just this `const_string`, but it's just a side information, it's also stored as `pattern`; this way the existing intersection logic still works

mmoskal: but actually, we can just get rid of const_string completely, and *always* const_compile the ::String!

mmoskal: it should be probably regex_compile not const_compile

hudson-ai:
Superceded by #100

Issue #87, opened 2024-12-10
implement intersection of pattern and min/max length


mmoskal:
We have machinery for this; covers 166 testcases

@hudson-ai not assigning in case I or someone else wants to do it but feel free, see also previous issues in this repo

mmoskal:
also regex&format and regex&regex will cover 12 more

hudson-ai:
As intersection happens in `schema.rs`, we don't yet have access to a `GrammarBuilder`. Do you think it makes sense to build derivre `RegexAst`s and use `GrammarBuilder.regex` later to add all the nodes, or do you think I should add a `RegexBuilder` to the `SharedContext` and just present the `GrammarBuilder` with a compiled regex when we get there?

Note: this question is getting at some of the (in my opinion) sharp edges w.r.t compositionality of the builders

hudson-ai:
Think I have a reasonable approach to my comment -- doing the former (building an ast and giving `RegexBuilder` an add_ast method).

Currently running into a different issue though -- json quoting currently doesn't support `And` nodes -- @mmoskal is this a "not going to implement" kind of thing, or can I add it to my wishlist? :)

mmoskal:
As far as I can tell I was just being too cautious. I think just removing the error checking should work for both `And` and `Not`. I don't see how they would be any different than `Or`. 
https://github.com/microsoft/derivre/commit/8a7e7224869e12b0bf316db29b9c5a3c8fae761c

hudson-ai:
Closed by #92

PR #92, opened 2024-12-13
Intersect patterns, format, and min/maxLength

hudson-ai:
@mmoskal a few things here
1. I added a `add_ast` method to the `RegexBuilder` that will traverse a whole `derivre::RegexAst` and add it to the builder. This feels a little clunky, so I'd appreciate your thoughts.
2. I parsed the `pattern` and `format` into `RegexAst`s in `schema.rs` and intersected them early. `minLength` and `maxLength` did not get this treatment, as taking maximums/minimums of numbers feels like a cleaner way to intersect these.
3. To intersect `minLength` and `maxLength` with a regex, I had to do something mirroring what I did in my other open PR, namely using `"(?s:.{{{},{}}})"` with json-quoting rather than `"\"{}{{{},{}}}\""` with `CHAR_REGEX`.
4. The current state of this PR preserves all existing behaviors, but it does NOT yet actually allow intersecting different types of string constraints... This is because json-quoting currently doesn't work with `And` (mentioned in a comment in #87)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -377,7 +377,7 @@ dependencies = [
 [[package]]
 name = "derivre"
 version = "0.1.0"
-source = "git+https://github.com/microsoft/derivre?rev=e03c91e5cfca55876eeaaecb845d4556f0222352#e03c91e5cfca55876eeaaecb845d4556f0222352"
+source = "git+https://github.com/microsoft/derivre?rev=8a7e7224869e12b0bf316db29b9c5a3c8fae761c#8a7e7224869e12b0bf316db29b9c5a3c8fae761c"
 dependencies = [
  "ahash",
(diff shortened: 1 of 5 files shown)


mmoskal:
I guess the `RegexAst` is more convenient to build since it's recursive, while the `RegexNode` requires a builder to store recursive nodes, and you don't have a builder when compiling `Value` -> `Schema`.

I think the solution you have is good enough. I don't think there's a point creating yet another type, which is like RegexAst but without ExprRef...

So good to go IMHO, and I fixed the derivre issue

hudson-ai:
Failure in guidance here is due to an explicit test that asserts that we raise an exception in this case -- will fix the test when I open a PR on guidance to bump the llg version

Issue #77, opened 2024-11-29
JSON: convert oneOf -> anyOf when safe


mmoskal:
Not urgent, but logging it for tracking:
- we currently have an option to approximate all `oneOf` as `anyOf`, which is unsound
- instead, we could check for a given `oneOf` if it is always equivalent to `anyOf`, in particular if it's something like an enum, or it has a discriminator field

hudson-ai:
Resolved by #94 , although additional cases can probably be marked safe

PR #94, opened 2024-12-13
Safe `oneOf`

hudson-ai:
Adds a method to check for "verifiable" schema disjointness (returns false if we're not really sure) and uses it to coerce oneOf to anyOf whenever it is provably safe to do so. This includes cases where objects have required properties that are themselves provably disjoint, covering discriminated unions.

We could probably take this one step farther and actually compute the difference between some schemas (e.g. `Schema::Boolean.subtract(Schema::LiteralBoolean{ true }) => Schema::LiteralBoolean{ false }`), but simply checking for disjointness covers some important cases.

Closes (or at least contributes significant progress towards) #77

### DIFF
--- a/parser/src/json/schema.rs
+++ b/parser/src/json/schema.rs
@@ -197,6 +197,175 @@ impl Schema {
             other_schema => other_schema,
         }
     }
+
+    /// Intersect two schemas, returning a new (normalized) schema that represents the intersection of the two.
+    fn intersect(self, other: Schema, ctx: &Context) -> Result<Schema> {
+        ctx.increment()?;
+
+        let merged = match (self, other) {
+            (Schema::Any, schema1) => schema1,
+            (schema0, Schema::Any) => schema0,
+            (Schema::Unsatisfiable { reason }, _) => Schema::Unsatisfiable { reason },
+            (_, Schema::Unsatisfiable { reason }) => Schema::Unsatisfiable { reason },
+            (Schema::Ref { uri }, schema1) => intersect_ref(ctx, &uri, schema1, true)?,
+            (schema0, Schema::Ref { uri }) => intersect_ref(ctx, &uri, schema0, false)?,
+            (Schema::OneOf { options }, schema1) => Schema::OneOf {
+                options: options
+                    .into_iter()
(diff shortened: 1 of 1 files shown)


hudson-ai:
This is fun.

This compiles:
```json
{"oneOf": [{"enum": [1,2,"foo"]}, {"enum": [2,3,"bar"]}], "type": "string"}
```
This does not:
```json
{"oneOf": [{"enum": [1,2,"foo"]}, {"enum": [2,3,"bar"]}]}
```

parser/src/json/schema.rs
```diff
@@ -192,11 +192,252 @@ impl Schema {
...
+                    .get(key)
+                    .unwrap_or(add1.as_deref().unwrap_or(&Schema::Any));
+                let prop2 = props2
+                    .get(key)
+                    .unwrap_or(add2.as_deref().unwrap_or(&Schema::Any));
+                prop1.is_verifiably_disjoint_from(prop2)
+            }),
+            _ => {
+                // Except for in the cases above, it should suffice to check that the types are different
+                mem::discriminant(self) != mem::discriminant(other)
```
mmoskal, line 437: Ha! I didn't know this existed!

Actually, on that note, maybe in future, Schema::Boolean should have value: Optional<bool> and we could drop LiteralBool? The same way numbers and strings are treated.

mmoskal (approved):
Looks great!

parser/src/json/schema.rs
hudson-ai: Oh, duh. That's a good idea. 😂

hudson-ai: @mmoskal I think we really need to implement refs here (could be in the next PR), but we'll definitely have to be careful about ref cycles causing infinite loops. Maybe setting a max ref depth on this method would be an ok solution

hudson-ai: P.s. matching on RegexAst::Literal can replace the const_string check in your const_compile PR.

mmoskal: I think we don't need to match on const anything - we just always compile strings as regexes and or them together with other options in AnyOf

Issue #84, opened 2024-12-09
compile enums as regexes, not grammars


mmoskal:
We should try to compile enums into a regex if possible.

Size distribution of enum sizes below - there is on the order of 100 cases with more than 200 enum items (which will blow the parser), and even for smaller ones regexes will be faster.

| Enum size | Num. schemas |
| ---- | ----- |
| 1    | 1920 |
| 2    | 1293 |
| 3-4    | 1763 |
| 5-8    | 957 |
| 9-16   | 444 |
| 17-32   | 390 |
| 33-64   | 263 |
| 65-128  | 88 |
| 129-256  | 51 |
| 257-512  | 15 |
| 513-1024 | 8 |
| 1025-2048    | 2 |
| 2049-4096    | 6 |
| 4097-8192    | 1 |

hudson-ai:
@mmoskal do you know if these are mostly string enums?

mmoskal:
among the non-synthesized cases:

```
  "enum:string_or_null": 3415,
  "enum:number_or_null": 143,
  "enum:boolean_or_null": 60,
  "enum:primitive": 44,
  "enum:mixed": 8
```


Where `primitive` means other combination of string, number, boolean, null.

The `mixed` cases are all weird.

hudson-ai:
just to do something quick and pragmatic -- it would be easy to rewrite the string enums as Schema::String variants with an alternation for the pattern. We could allow nullability by wrapping that in a Schema::OneOf -- would this take care of most or all of the "large" enums?

mmoskal:
I would add `is_const()` method `Schema`, plus a `const_val: Option<String>` to `Schema::String`; then additional method `const_compile() -> Option<String>`; then when compiling AnyOf you check if they are all const and if so map with const_compile() and do an OR

mmoskal:
Couldn't help myself #91 ;)

hudson-ai:
@mmoskal good man!! I'll take a look later :)

PR #100, opened 2024-12-23
Use RegexAst::Literal instead of const field on strings for fast enums

hudson-ai:


hudson-ai:
@mmoskal kind of messed up the base of this pr as I had to merge main back into yours

parser/src/json/schema.rs
```diff
@@ -126,6 +128,43 @@ impl Schema {
...
+                maximum: Some(y),
+                ..
+            } if x == y => return Some(RegexNode::Literal(x.to_string())),
+            Schema::String {
+                // TODO: check that min/max don't break the literal and all the other constraints
+                // or just convert this whole node to a regex
+                min_length: 0,
+                max_length: None,
+                regex: Some(RegexAst::Literal(s)),
+            } => &regex_syntax::escape(&serde_json::to_string(s).unwrap()),
```
hudson-ai, line 146: This is the bit that makes it a bit tricky to do this for general string regexes -- json quoting

mmoskal: I'll add AST node to escape. Shouldn't be hard.

hudson-ai:
Closes #84

PR #101, opened 2025-01-02
compile strings etc as regexes

mmoskal:
cc @hudson-ai for awareness; still WIP!

### DIFF
--- a/parser/src/json/compiler.rs
+++ b/parser/src/json/compiler.rs
@@ -608,6 +608,14 @@ impl Compiler {
             return Ok(self.json_quote(RegexAst::Regex("(?s:.*)".to_string())));
         }
         if let Some(mut ast) = regex {
+            let mut positive = false;
+
+            fn mk_rx_repr(ast: &RegexAst) -> String {
+                let mut rx_repr = String::new();
+                ast.write_to_str(&mut rx_repr, 1_000, None);
+                rx_repr
+            }
+
             if min_length != 0 || max_length.is_some() {
                 ast = RegexAst::And(vec![
                     ast,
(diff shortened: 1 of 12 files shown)


mmoskal:
I think this is good to go. It happens. to fix an enum quoting issue that's present on `main` (see added testcase)

hudson-ai:
> I think this is good to go. It happens. to fix an enum quoting issue that's present on `main` (see added testcase)

Awesome, will try to review today!

parser/src/json/compiler.rs
```diff
@@ -531,12 +502,98 @@ impl Compiler {
...
+            Schema::Null => literal_regex("null"),
+            Schema::Boolean => Some(RegexAst::Regex("true|false".to_string())),
+            Schema::LiteralBool { value } => literal_regex(if *value { "true" } else { "false" }),
+
+            // TODO: is this special case needed?
+            Schema::Number {
+                minimum: Some(x),
+                maximum: Some(y),
+                ..
+            } if x == y => Some(RegexAst::Literal(x.to_string())),
```
hudson-ai, line 530: Can this special case actually get us into trouble, e.g. if min == max, but exclusiveMinimum > minimum, or if min == max is a non-integer, but we have integer=True?

Btw I'm open to changing our representation to exclusive{Minumum,Maximum} being bools if it simplifies things.

hudson-ai:
Bar my one comment, lgtm!

parser/src/json/compiler.rs
mmoskal: Let's nuke it then

hudson-ai (approved):

PR #105, opened 2025-01-07
[Draft] Add JSON Schema Test Suite to CI

hudson-ai:
First attempt at getting the test suite running in CI. I don't really know how to effectively use github CI, so this is a bit of an experiment.

Note that this steals some code from a few existing scripts (namely `sample_parser/src/json_schema_testsuite.rs`).

TODOS:
- [ ] commit some baseline version of the artifact for regression analysis
- [ ] compare against baseline with helpful presentation of differences
- [ ] build some process for bumping the baseline (probably just run a script and update committed baseline)


mmoskal:
You could just add it to the test-guidance.sh script (which we probably want to rename to build-and-test.sh or testall.sh or something). This way there is just one script to run to check if things still work locally. I guess running this suite takes a few seconds at most?

hudson-ai:
> You could just add it to the test-guidance.sh script (which we probably want to rename to build-and-test.sh or testall.sh or something). This way there is just one script to run to check if things still work locally. I guess running this suite takes a few seconds at most?

Sounds good, and yeah it's under a second :)

mmoskal:
@hudson-ai please reopen if we still want to do it

PR #106, opened 2025-01-08
[JSON] Allow custom retrievers

hudson-ai:
Not high priority, but I think it's reasonable to allow users to pass custom retrievers if they want to resolve remote refs (we should NOT provide a default implementation of this though).

This also allows us to try more cases in the test suite, as we can just put all of the remote refs that the tests depend on into a HashMap which we thinly wrap with a Retrieve implementation.

I am very new to all this `dyn Trait` stuff, so I'm guessing I did something silly.

### DIFF
--- a/parser/src/json/schema.rs
+++ b/parser/src/json/schema.rs
@@ -2,8 +2,9 @@ use anyhow::{anyhow, bail, Result};
 use derivre::RegexAst;
 use hashbrown::{HashMap, HashSet};
 use indexmap::{IndexMap, IndexSet};
-use referencing::{Draft, Registry, Resolver, ResourceRef};
+use referencing::{Draft, Registry, Resolver, Resource, ResourceRef, Retrieve};
 use serde_json::Value;
+use std::any::type_name_of_val;
 use std::{cell::RefCell, mem, rc::Rc};
 
 use super::formats::lookup_format;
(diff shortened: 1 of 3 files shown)


parser/src/json/schema.rs
```diff
@@ -567,7 +589,18 @@ pub fn build_schema(contents: Value) -> Result<(Schema, HashMap<String, Schema>)
...
-    let registry = Registry::try_new(&base_uri, resource)?;
+    let registry = {
+        // Weirdly no apparent way to instantiate a new registry with a retriever, so we need to
+        // make an empty one and then add the retriever + resource that may depend on said retriever
+        let empty_registry =
+            Registry::try_from_resources(std::iter::empty::<(String, Resource)>())?;
+        let retriever = if let Some(wrapper) = retriever {
+            wrapper.as_ref()
+        } else {
+            &referencing::DefaultRetriever
```
hudson-ai, line 600: Note that this is effectively the NULL retriever, as its implementation of `retrieve` is to just return an error.

mmoskal (commented):
LGMT! I'm not sure what that Deref stuff is but if it's working, it's OK :)

hudson-ai:
> I'm not sure what that Deref stuff is

That certainly doesn't inspire a ton of confidence in my approach, but hey -- if you're cool with it!

hudson-ai:
Note: the wrapper literally only exists to implement Debug. I initially tried using a `Rc<dyn Retrieve + Debug>` in the compile options, which turned into having to implement a new trait, which snowballed into weird visibility issues with the trait when used by an external crate, and I gave up... This corner of rust still feels a little spooky to me.

hudson-ai:
@mmoskal side note, but someday we may want to expose python bindings for this. Will be interesting to learn how to do that.

mmoskal:
It's normal you can't implement traits for structs from outside crates. You always have to do a wrapper in this case (unless it's a trait from your crate).

PR #116, opened 2025-02-03
[Draft] Port over guidance's 'substring'

hudson-ai:
TODOs:
- [ ] provide public/python interface
- [ ] add substring to grammar rather than just building a derivre AST
- [ ] remove `regex` dependency and/or offer more flexible "tokenization"


hudson-ai:
@mmoskal ping for visibility -- I don't have a whole lot of time to work on this at the moment, but I figured getting it in front of you sooner rather than later would be good!

hudson-ai:
From @mmoskal
> API should be substring(chunks: list[str]) not substring(s: str, words: bool) - this way it's more flexible and you get rid of rust regex dep; of course Python can have default split on words or whatever

I agree

PR #119, opened 2025-02-12
Allow passing string for `capture_name` in lark syntax

hudson-ai:
allow passing string for capture_name to allow keys that may be illegal rule names


mmoskal:
Maybe just keep the old syntax as well? Used in tests and shorter.

hudson-ai:
> Maybe just keep the old syntax as well? Used in tests and shorter.

Fair -- put it back! 😄

hudson-ai:
Even better. Thanks @mmoskal!

PR #128, opened 2025-02-24
support for INDENT/DEDENT tokens

### LINKED ISSUE #107: Indent/Dedent
- For JSON and Python: track indentation depth
- For Python: also track nesting level of parentheses

mmoskal:
Fixes #107

### DIFF
--- a/parser/src/api.rs
+++ b/parser/src/api.rs
@@ -3,6 +3,7 @@ use std::{
     ops::RangeInclusive,
 };
 
+use anyhow::{ensure, Result};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use toktrie::TokenId;
@@ -47,13 +48,27 @@ pub struct LLGuidanceOptions {
 }
 
 impl LLGuidanceOptions {
-    pub fn apply(&mut self, other: &LLGuidanceOptions) {
+    pub fn apply(&mut self, other: &LLGuidanceOptions) -> Result<()> {
         if other.no_forcing {
             self.no_forcing = true;
         }
         if other.allow_invalid_utf8 {
             self.allow_invalid_utf8 = true;
         }
+        if self.indent.is_none() {
+            self.indent = other.indent.clone();
+        } else {
+            ensure!(self.indent == other.indent, "Incompatible indent options");
+        }
+        if self.indent_newline_rx.is_none() {
+            self.indent_newline_rx = other.indent_newline_rx.clone();
+        } else {
+            ensure!(
+                self.indent_newline_rx == other.indent_newline_rx,
+                "Incompatible indent_newline_rx options"
+            );
+        }
+        Ok(())
     }
 }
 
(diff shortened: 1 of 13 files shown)


nchammas:
Hello @mmoskal. I am trying to have Guidance generate code with significant indentation, so I am following your work here.

Do you intend as part of this work to implement support for Lark's `declare` statement? It's a critical part of how Lark enables support for significant indentation, at least as far as I understand it from reviewing [the relevant docs][1] as well as [Lark's Python grammar][2].

[1]: https://lark-parser.readthedocs.io/en/latest/examples/indented_tree.html
[2]: https://github.com/lark-parser/lark/blob/9a125777b79f551e06aa368a86d61d89f0057019/lark/grammars/python.lark#L275

mmoskal:
`%declare` in Lark just says the definition of the token is provided elsewhere. For constraint we actually need to know what `IDENT` and `DEDENT` tokens do. Unfortunately, it's far from simple.

nchammas:
Yes, in Lark you need to provide an instance of `lark.indenter.Indenter`:

```python
class TreeIndenter(Indenter):
    NL_type = '_NL'
    OPEN_PAREN_types = []
    CLOSE_PAREN_types = []
    INDENT_type = '_INDENT'
    DEDENT_type = '_DEDENT'
    tab_len = 8

parser = Lark(tree_grammar, parser='lalr', postlex=TreeIndenter())
```

I already have an indentation-significant DSL built using Lark, and I am hoping to use the same grammar mostly as-is with Guidance to have an LLM output queries in my DSL. To do that, Guidance's Lark interface would probably need to accept some bit of configuration equivalent to the above.

Is that something you are planning to do, or will the approach be very different? I know this is a work in progress, so I don't expect any definite answers. Just sharing my use case.

Separately, would it help at all if Lark itself provided some kind of API to help with next token prediction? I can see that you have built your own implementation of Lark in Rust (roughly speaking), but I wonder if direct support from Lark itself would also be useful somehow.

mmoskal:
there is [a similar setup](https://github.com/guidance-ai/llguidance/pull/128/files#diff-292f2f86a0f85ef4e8d1251b553d30fcb16e2e59dca4ac69718807a65f2b7e6aR751) in this PR

When designing a new DSL to be written by LLMs I would suggest not using indentation. AFAIU it makes the LLM stupider, as it has to keep track of it, instead of simply following braces.

Unfortunately, any changes in Lark Python code cannot be used in LLGuidance.

nchammas:
> When designing a new DSL to be written by LLMs I would suggest not using indentation. AFAIU it makes the LLM stupider, as it has to keep track of it, instead of simply following braces.

Oh, that's surprising to hear. I assumed that due to the popularity of Python and YAML, LLMs wouldn't have a particular problem with indentation-significant languages. I wonder why tracking indentation level would be much harder for LLMs than tracking braces.

The DSL I designed is primarily for humans to write, but I am exploring how practical it is to have an LLM assist non-technical users by converting the queries they write in English into DSL queries.

If indentation is such a problem, perhaps I should develop some kind of additional JSON format for my DSL just for LLMs to target. Then I would be able to use the much more mature support for JSON schema-constrained generation. Not sure if this would be a lot of work, or if it would work well, as I haven't worked with JSON schemas before. But that's what I'll research next, I think.

PR #130, opened 2025-02-25
[JSON] "x-guidance" JsonCompileOptions

hudson-ai:
Exposes JsonCompileOptions under a key `x-guidance` in the top level of a JSON schema

### DIFF
--- a/parser/src/json/compiler.rs
+++ b/parser/src/json/compiler.rs
@@ -19,17 +19,13 @@ use crate::{
 // TODO: schemastore/src/schemas/json/BizTalkServerApplicationSchema.json - this breaks 1M fuel on lexer, why?!
 
 #[derive(Debug, Clone, Deserialize)]
-#[serde(deny_unknown_fields)]
+#[serde(default, deny_unknown_fields)]
 pub struct JsonCompileOptions {
-    #[serde(default)]
     pub item_separator: String,
-    #[serde(default)]
     pub key_separator: String,
-    #[serde(default)]
     pub whitespace_flexible: bool,
-    #[serde(default)]
     pub coerce_one_of: bool,
-    #[serde(default, skip_deserializing)]
+    #[serde(skip_deserializing)]
     pub retriever: Option<RetrieveWrapper>,
 }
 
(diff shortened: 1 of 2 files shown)


parser/src/json/compiler.rs
```diff
@@ -17,12 +18,17 @@ use crate::{
 // TODO: array maxItems etc limits
 // TODO: schemastore/src/schemas/json/BizTalkServerApplicationSchema.json - this breaks 1M fuel on lexer, why?!
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Deserialize)]
```
mmoskal, line 21: do we want to:

```suggestion
#[derive(Debug, Clone, Deserialize)]
#[serde(deny_unknown_fields)]
```

hudson-ai: Not a bad idea!

parser/src/json/compiler.rs
```diff
@@ -17,12 +18,17 @@ use crate::{
...
 pub struct JsonCompileOptions {
+    #[serde(default)]
```
mmoskal, line 24: actually, this will default the field to `""` not to the default from `::default`; I think you need to make these fields `Option<String>`

mmoskal (changes requested):

parser/src/json/compiler.rs
hudson-ai: Ahhhhh got it. Will do

hudson-ai: i.e. like 
```
#[derive(Debug, Clone, Deserialize)]
#[serde(default, deny_unknown_fields)]
pub struct JsonCompileOptions {
    pub item_separator: String,
    pub key_separator: String,
    pub whitespace_flexible: bool,
    pub coerce_one_of: bool,
    #[serde(skip_deserializing)]
    pub retriever: Option<RetrieveWrapper>,
}
```

hudson-ai: https://serde.rs/container-attrs.html#default
seems to be my intended behavior :)

mmoskal: TIL!

mmoskal (approved):

PR #137, opened 2025-03-13
Check for presence of any *keyword* that is not listed in META_AND_ANNOTATIONS, not any *key*

### LINKED ISSUE #135: dummy keywords in schemas
A schema `{ "title": "something" }` and `{ "x-title": "something" }` is treated differently due to this function:

```rust
fn only_meta_and_annotations(schemadict: &HashMap<&str, &Value>) -> bool {
    schemadict.keys().all(|k| META_AND_ANNOTATIONS.contains(k))
}
```

I think it should check for presence of any *keyword* that is not listed in `META_AND_ANNOTATIONS`, not any key.

Now, this may be pointing at another issue: If you  add `"x-foo":"bar"` to `Github_medium---o61004.json`, then the `oneOf` will be compiled OK, but without it we bail.

hudson-ai:
Fixes #135

### DIFF
--- a/parser/src/json/schema.rs
+++ b/parser/src/json/schema.rs
@@ -509,15 +509,14 @@ fn compile_contents_inner(ctx: &Context, contents: &Value) -> Result<Schema> {
     compile_contents_map(ctx, schemadict)
 }
 
-fn only_meta_and_annotations(schemadict: &HashMap<&str, &Value>) -> bool {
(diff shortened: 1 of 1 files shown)


mmoskal (approved):

mmoskal:
@hudson-ai can you merge main here, so it re-runs the tests?

PR #139, opened 2025-03-14
chore(lib): add py.typed for annotations information

aarnphm:
https://peps.python.org/pep-0561/ to determine that llguidance is fully typed.

This will help mypy to infer library type accordingly, to avoid https://github.com/vllm-project/vllm/pull/14589/files `type: ignore[import-untyped]`

Signed-off-by: Aaron Pham <contact@aarnphm.xyz>


mmoskal:
Thanks! Trying to setup mypy in CI now

PR #172, opened 2025-05-06
use fancy-regex instead of onig as tokenizers regex library

Slowki:
The version of Oniguruma used in `onig_sys` doesn't build on GCC 15 and the oniguruma project itself got archived last week, so this PR switches tokenizers to the fancy-regex backend.

`fancy-regex` also requires flipping on the `unstable_wasm` feature until huggingface/tokenizers#1772 lands, that flag doesn't have any ill effects though since everything WASM related downstream is behind `target_arch` checks.

**tl;dr**: This fixes builds on Linux distros with newer GCC versions like Arch Linux and Fedora.


mmoskal (approved):
Thank you, looks good!

PR #175, opened 2025-05-19
Stop using prefix/sentinel strings for `TokenizerWrapper`

hudson-ai:
Make it the responsibility of the caller (`guidance`) to ensure proper tokenization under boundary effects (e.g. by removing leading space added by sentencepiece )

### DIFF
--- a/python/llguidance/_tokenizer.py
+++ b/python/llguidance/_tokenizer.py
@@ -23,10 +23,6 @@ class TokenizerWrapper:
             gtokenizer(b"test")
         except:
             self._accepts_bytes = False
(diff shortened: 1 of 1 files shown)


mmoskal (approved):

mmoskal:
@hudson-ai let me know if you want me to do a release with this (should have merged earlier today...)

hudson-ai:
@mmoskal I think this is okay to wait until the next natural release unless @lochuynh1412 / @marcotcr runs into more issues with the gemma tokenizer and `\x02`.

I think that the issue they were running into was likely due to the sentinel bytes being used in both libraries. Otherwise, they would only see such an issue when the model actually produces a token starting with `\x02`, which I imagine is rather rare.

PR #181, opened 2025-05-27
add parse_special=False to tokenize_str/bytes() in python

mmoskal:
@hudson-ai does this look good for your purposes?


hudson-ai:
Looks perfect, thank you for the quick turnaround! :)

PR #192, opened 2025-06-18
parametric grammars

### LINKED ISSUE #191: parametric grammars
- add a 64-bit parameter to symbol names
- allow rules conditional on bits being set in the parameter
- allow expressions in rules to set parameters to RHS symbols
- useful for specifying "unique", "contains" constraints in JSON schema and similar

Example - permutation of N elements:

```lark
start: list#(0)
list#(m if m & (1 << k) == 0): elt_k list#(m | (1 << k)) // repeat for k in 0..N
list#(m if m == (1 << N) - 1): ""
elt_0: ...
elt_1: ...
...
elt_N-1: ...
```

Example, any sequence of N elements, where each has to occur at least once:

```lark
start: list#(0)
list#(m if m & (1 << k) == 0): others#(m) elt_k list#(m | (1 << k)) // repeat for k in 0..N
list#(m if m == (1 << N) - 1): others#(m)
others#(m): others#(m) other#(m) | ""
others#(m if m & (1 << k) != 0): elt_k // repeat for k in 0..N
```

Expressions:

```
add_bit(m, k) = m | (1 << k)
del_bit(m, k) = m & ~(k << k)
is_set(m, k) = m & (1 << k) != 0
is_clear(m, k) = m & (1 << k) == 0
is_all_set(m, k) = m == (1 << (n+1)) - 1
bit(k) = 1 << k
all_bits(n) = (1 << (n+1)) - 1
empty() = 0
```

```lark
start: list#(empty())
list#(m if is_clear(m, k)): others#(m) elt_k list#(add_bit(m, k)) // repeat for k in 0..N
list#(m if is_all_set(m, N)): others#(m)
others#(m): others#(m) other#(m) | ""
others#(m if is_set(m, k)): elt_k // repeat for k in 0..N
```

mmoskal:
fixes #191

### DIFF
--- a/docs/parametric.md
+++ b/docs/parametric.md
@@ -1,0 +1,48 @@
+# Parametric grammars
+
+- add a 64-bit parameter to symbol names
+- allow rules conditional on bits being set in the parameter
+- allow expressions in rules to set parameters to RHS symbols
+- useful for specifying "unique", "contains" constraints in JSON schema and similar
+
+Example - permutation of N elements:
+
+```lark
+start: list#(0)
+list#(m if m & (1 << k) == 0): elt_k list#(m | (1 << k)) // repeat for k in 0..N
+list#(m if m == (1 << N) - 1): ""
+elt_0: ...
+elt_1: ...
+...
+elt_N-1: ...
+```
+
+Example, any sequence of N elements, where each has to occur at least once:
+
+```lark
+start: list#(0)
+list#(m if m & (1 << k) == 0): others#(m) elt_k list#(m | (1 << k)) // repeat for k in 0..N
+list#(m if m == (1 << N) - 1): others#(m)
+others#(m): others#(m) other#(m) | ""
+others#(m if m & (1 << k) != 0): elt_k // repeat for k in 0..N
+```
+
+Expressions:
+
+```
+insert(m, k) = m | (1 << k)
+remove(m, k) = m & ~(1 << k)
+has(m, k) = m & (1 << k) != 0
+has_not(m, k) = m & (1 << k) == 0
+has_all(m, n) = m == (1 << n) - 1
+set(k0,...,kN) = (1 << k0) | ... | (1 << kN)
+full(n) = (1 << (n+1)) - 1
+```
+
+```lark
+start: list#(set())
+list#(m if has_not(m, k)): others#(m) elt_k list#(insert(m, k)) // repeat for k in 0..N
+list#(m if has_all(m, N)): others#(m)
+others#(m): others#(m) other#(m) | ""
+others#(m if has(m, k)): elt_k // repeat for k in 0..N
+```
--- a/parser/src/earley/grammar.rs
+++ b/parser/src/earley/grammar.rs
@@ -29,6 +29,13 @@ impl Symbol {
             self.name.clone()
         }
     }
+    fn param_name(&self, param: &ParamExpr) -> String {
+        let mut r = self.short_name();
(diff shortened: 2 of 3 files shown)


hudson-ai:
Am I understanding correctly that this should let us describe some grammars like (permutations of N) in a linear # of rules rather than exponential? That's darn cool!

And just to clarify, do you know if this is still context free (I think yes)? It does sort of look effectively S-attributed though (if I understand correectly what that means).

mmoskal:
@hudson-ai yes, you can do permutations see first example in #191 

It is technically context-free but with huge number of rules (I allows for 64 bits of context; we could do more I guess).

It's not S-attributed; it's more like the "templates" in Lark that define a family of rules. Still thinking about the syntax.

mmoskal:
Permutation of a0,a1,a2:

```lark
start: perm{set()}
perm{#m}:   ""                        %if has_all(m, 3)
        |   a0 perm{insert(m, 0)}     %if has_not(m, 0)
        |   a1 perm{insert(m, 1)}     %if has_not(m, 1)
        |   a2 perm{insert(m, 2)}     %if has_not(m, 2)
```

Sequence with at least one each of a0, a1, a2:

```lark
start: list{set()}
list{#m}:   other{m}*                        %if has_all(m, 3)
        |   other{m}* a0 list{insert(m, 0)}  %if has_not(m, 0)
        |   other{m}* a1 list{insert(m, 1)}  %if has_not(m, 1)
        |   other{m}* a2 list{insert(m, 2)}  %if has_not(m, 2)
other{#m}: a0   %if has(m, 0)
         | a1   %if has(m, 1)
         | a2   %if has(m, 2)
```

hudson-ai:
Thoughts on supporting multiple arguments? Might make certain ideas easier to express, especially if there are multiple orthogonal things to track that are inconvenient to pack into a single u64. (Disregard if you already support this)

mmoskal:
@hudson-ai we can possibly do multi-arguments in future - internally I think it's a matter of supporting operators on bit ranges

If you have ideas what to use this for, other than permutations/uniqueness, please post them here (in particular if you need more args or additional operators).

I was thinking of grammar `a* b*` where the length of the whole grammar is limited to some number (each `a` or `b` count as 1) - this is an extraction of some problem we have discussed in the past but I'm not sure which one. I'm also not sure we can do this with the parametric grammars as they stand now (at least we would need addition and less-than).

hudson-ai:
Not sure what else we'd want besides these examples! Need to think a bit.

Still in the vein of permutations, but choosing k elts uniquely from a set of n (pseudocode):
```lark
start: n_c_k{set()}
n_c_k{#m}:   ""                      %if ge(bit_count(m), k)
        |   a0 n_c_k{insert(m, 0)}  %if has_not(m, 0) & lt(bit_count(m), k)
        |   a1 n_c_k{insert(m, 1)}  %if has_not(m, 1) & lt(bit_count(m), k)
        ...
```

mmoskal:
@hudson-ai @Harsha-Nori @marcotcr @slundberg 

Please take a look (on Monday I guess ;) - in particular look at the [new docs](https://github.com/guidance-ai/llguidance/blob/parametric/docs/parametric.md) and see if there is something missing - feel free to ask "how do I do X with it"

(I may merge it earlier but we can always iterate)

hudson-ai:
Good stuff! And happy 1.0.0!

Will definitely be pinging you as I add some JSON features that take advantage of this... And if/when I think of other interesting use cases. Thanks @mmoskal

PR #194, opened 2025-06-28
Strip debug info from the wheels

Ahajha:
This greatly shrinks the final wheels, from ~56 MB to ~9.5 MB.

Originally we tried full stripping, that would bring the wheels down to ~7.5 MB. Just symbols seems like a reasonable middle ground.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ resolver = "2"
 
 [profile.release]
 debug = 1
+strip = "debuginfo"
 # lto = "thin"
 # codegen-units = 1
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,5 +24,4 @@ python-source = "python"
 module-name = "llguidance._lib"
 manifest-path = "python_ext/Cargo.toml"
 sdist-generator = "git"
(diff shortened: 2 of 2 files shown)


mmoskal:
Generally, the reason not to strip them is that the crash reports are much more informative. Maybe it's possible to somehow keep the debug info on the side.

Ahajha:
It might mean an extra step, but in the worst case a user could rerun with a debug-enabled binary after, the presence of debug symbols shouldn't affect that. Do you get many of these reports? It's Rust, so I would generally assume crashes are less frequent than in other languages :)

Ahajha:
I wonder if there are any options to compress the debug info, I might poke around.

mmoskal:
["-C", "strip=debuginfo"] should help but not yet sure where to pass it

Ahajha:
@mmoskal I updated to only strip debug info, this brings the size to 9.48 MB (x86 Linux still), which I think is reasonable, now it's ~2 MB for debug symbols instead of ~45 for full debug info. WDYT?

mmoskal:
Thank you!

PR #195, opened 2025-06-28
fix: tokenizers normalizers sequence api changed

ammar-elsabe:
4 days ago huggingface released `tokenizers@v0.21.2`, which [changed the `get_normalizers` method to be an `impl AsRef<[NormalizerWrapper]> for Sequence`](https://github.com/huggingface/tokenizers/commit/c45aebd1029acfbe9e5dfe64e8b8441d9fae727a#diff-252dff06be81a9c396077c0b02f9f6aef6772058348bd6584031a48b351d623e). Due to the version requirement in `toktrie_hf_tokenizers` (`version = ">=0.20.0, <1.0.0"`), in the absence of a lockfile cargo fetches v0.21.2 which will not compile

This was caught in EricLBuehler/mistral.rs#1523

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,22 +13,22 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "ahash"
-version = "0.8.11"
+version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
+checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
 dependencies = [
  "cfg-if",
- "getrandom 0.2.15",
+ "getrandom 0.3.3",
  "once_cell",
  "serde",
  "version_check",
- "zerocopy 0.7.35",
+ "zerocopy",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


ammar-elsabe:
I didn't include this in the PR but if huggingface is going to introduce api breaking changes with patches maybe the version constraint should be `=0.21.2`?

mmoskal:
The main reason I had this very flexible dep is so that the tokenizer lib can be easily updated in mistral.rs (I think that's the only pure rust user of llg). But I think you're right - if they break it like this, we should probably pin it...

ammar-elsabe:
Maybe make it "0.21.2" and hope they follow semver guidelines from now on?

dirvine:
This would be really helpful if we could progress it.

hudson-ai:
> Maybe make it "0.21.2" and hope they follow semver guidelines from now on?

@ammar-elsabe did you mean to set it to `version = ">=0.21.2, <1.0.0"`? While my opinion on this isn't *incredibly* strong, I do lean a bit towards us trying to remain flexible by bumping the lower-end of the range and optimistically hoping they get a bit better at semver... If this bites us again, I'll be more on the "pin it" team.

mmoskal:
I'm with @hudson-ai on this one - please set 0.21.2 as minimum and let's see!

ammar-elsabe:
@hudson-ai I actually set it to "0.21.2" on purpose, technically speaking SemVer [considers any version change before 1.0.0 to be incompatible](https://semver.org/#spec-item-5)

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

cargo is a little more flexible, [considering minor version bumps as incompatible pre-1.0.0 but patches are allowed](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#default-requirements)

> Versions are considered compatible if their left-most non-zero major/minor/patch component is the same. This is different from SemVer which considers all pre-1.0.0 packages to be incompatible.

I thought `0.21.2 := >=0.21.2, <0.22.0` would be a decent middleground that follows cargo's guidelines, regardless its your repository so up to you :shrug: , ill update it with whatever you confirm

hudson-ai (approved):
Ah, I wasn't aware of cargo's flexibility on patch versions for pre-`1.0.0`. This means `0.21.2 := >=0.21.2, <0.22.0`? I would agree that this is a reasonable middle-ground. @mmoskal what do you think?

mmoskal (approved):

Issue #200, opened 2025-07-10
how to use /.*/  represent the range of all tokens


ZonePG:
If I want to use /.*/ to represent the range of all tokens in the vocabulary instead of directly using <[0-vocabsize]>, where should I start?

Because I want to use such syntax in the terminal:

mygen[suffix='</think>] = <[0-155135]>*

155136 is my tokenizer vocab size.

hudson-ai:
`<[*]>` should now work to denote the range of all tokens 😄 . Please reopen if it gives you trouble!

PR #201, opened 2025-07-10
Add lark syntax for "any token" and negation of token ranges

hudson-ai:
Proposes a syntax (`<[*]>`) to address #200, furthermore introduces token-class complements following semantics of regex character classes.

### DIFF
--- a/parser/src/grammar_builder.rs
+++ b/parser/src/grammar_builder.rs
@@ -375,17 +375,18 @@ impl GrammarBuilder {
 
     pub fn any_token(&mut self) -> Result<NodeRef> {
         self.check_limits()?;
-        let trie = self
-            .tok_env
-            .as_ref()
-            .ok_or_else(|| anyhow!("no tokenizer - can't validate <any_token>"))?
-            .tok_trie();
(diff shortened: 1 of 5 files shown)


hudson-ai:
@lochuynh1412 fyi

hudson-ai:
Would like some feedback on the proposed syntax -- any objections or alternative suggestions @mmoskal ?

parser/src/grammar_builder.rs
```diff
@@ -296,10 +298,50 @@ impl GrammarBuilder {
             }
         }
 
+        let token_ranges = if negate {
+            let trie = if let Some(trie) = &trie {
+                trie
+            } else {
+                bail!("negation of token ranges is not supported without a tokenizer");
```
mmoskal, line 305: No tokenizer is only possible if we're just validating the grammar, not actually compiling it. You can do the same as `special_token` below - return `INVALID_TOKEN` or similar and add warning

mmoskal:
Syntax looks good! Make sure to add docs as well.

mmoskal (approved):
Modulo typo

Issue #207, opened 2025-07-21
[Bug] Error message for unsatisfiable `integer` and `number` schema


riedgar-ms:
When the error message is generated for mixed up bounds, the exclusivity of the bound isn't being respected. For example
```rust
    json_err_test(
        &json!({
            "type": "integer",
            "exclusiveMinimum": 1, "exclusiveMaximum": -1
        }),
        "Unsatisfiable schema: minimum (1) is greater than maximum (-1)",
    );
```
actually passes.

riedgar-ms:
This may well be related to #206

Issue #206, opened 2025-07-21
[Bug] Error on unsatisfiable integer schema not consistent


riedgar-ms:
This is minor, but I've found as I'm writing tests:
```rust
    json_err_test(
        &json!({
            "type": "integer",
            "exclusiveMinimum": 1, "exclusiveMaximum": -1
        }),
        "Unsatisfiable schema",
    );
```
(and all other combinations of exclusive and non-exclusive), but
```rust
    json_err_test(
        &json!({
            "type": "integer",
            "exclusiveMinimum": 0, "exclusiveMaximum": 1
        }),
        "Failed to generate regex for integer range",
    );
```

PR #208, opened 2025-07-21
[Test] Start porting JSON tests

riedgar-ms:
Begin porting the JSON tests over from Python, since they really belong here.

Have had to add some clippy suppressions for reasons which escape me.

### DIFF
--- a/sample_parser/tests/common_lark_utils/mod.rs
+++ b/sample_parser/tests/common_lark_utils/mod.rs
@@ -29,6 +29,7 @@ pub fn consume(parser: &mut TokenParser, tok: u32) {
     assert!(n == 0);
 }
 
+#[allow(dead_code)]
 pub fn lark_ok(lark: &str) {
     if let Err(e) = make_parser(lark, false) {
         panic!("unexpected error: {e}, grm:\n{lark}")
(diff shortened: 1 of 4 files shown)


riedgar-ms:
Some points:
- I'd prefer to have a `json` subdirectory, but this appears to disagree with how `cargo` thinks Things Should Be
- I'm not a fan of the `json_test_many()` function, but that seems to be the best one can do with the built in testing functionality in `rust`. To get something closer to `pytest.parametrize()` we would need to use `rstest` (or something of that ilk)
- I've noticed a few inconsistencies around unsatisfiable schema, and have opened issues (#206 and #207). For now, the tests enforce current behaviour
- This is not a porting of everything we have in the Python tests. It's a start which leaves out a lot of the edge cases

mmoskal:
Feel free to add rstest if it helps.

riedgar-ms:
Since I'm still learning Rust, I'd rather merge this, and then review the available crates for testing. It looks like `rstest` will give something like what I'd like, but I've not done a thorough search.

Also @mmoskal , would you object if I were to get the ability to run workflows on my own PRs? I think that's 'Write' or 'Maintain' permissions.

Issue #211, opened 2025-07-23
request: Restore the `LLMatcher` when it consumes an unacceptable token


syuoni:
Hi llguidance developers,

Thanks for this great project. I'm working on the integration to TensorRT-LLM.

I found when passing an unacceptable token to  `LLMatcher.consume_token`, the matcher gets in the error state, and seems there is no way to restore it back to the state before calling `LLMatcher.consume_token`.

-----

I tried `LLMatcher.try_consume_tokens` and `LLMatcher.validate_tokens`, which look like that they can consume the acceptable tokens only, without getting the matcher into the error state. However, I found these two APIs can still make the matcher into a error state:

```python
from transformers import AutoTokenizer
import llguidance
import llguidance.hf

tokenizer = AutoTokenizer.from_pretrained("meta-llama/Llama-3.1-8B-Instruct")

schema = '{"title": "WirelessAccessPoint", "type": "object", "properties": {"ssid": {"title": "SSID", "type": "string"}, "securityProtocol": {"title": "SecurityProtocol", "type": "string"}, "bandwidth": {"title": "Bandwidth", "type": "string"}}, "required": ["ssid", "securityProtocol", "bandwidth"]}'


grammar = llguidance.LLMatcher.grammar_from_json_schema(schema)

ll_tokenizer = llguidance.hf.from_tokenizer(tokenizer)
matcher = llguidance.LLMatcher(ll_tokenizer, grammar)

for tid in [26356, 220, 330, 62843, 794, 330, 24861, 7099, 50913, 4828, 220, 330, 17476, 21346, 794, 330, 54, 8201, 17, 12, 86747, 4828, 220, 330, 7198, 3175, 794, 330, 5894, 15, 80575, 320, 20, 52405, 8, 5139, 92, 364, 364, 482]:
    matcher.try_consume_tokens([tid])
    if matcher.is_error():
        print(matcher.get_error())
```

hudson-ai:
Thanks for your work on the TRT-LLM integration! Really excited to get it working with speculative decoding :)

I can confirm the repro -- it looks like `try_consume_tokens` and `validate_tokens` both put the matcher into an error state if they are called after the matcher is "stopped", i.e. `matcher.is_stopped() == True`.

@mmoskal it seems that this is due to `check_initialized` asserting that the matcher is not in a stopped state. Would relaxing / removing that have any negative implications that you are aware of?

[...]

PR #212, opened 2025-07-23
[Bug] `matcher.try_consume_tokens` not checking stop if some tokens are left unconsumed

hudson-ai:
The bug: (found while debugging https://github.com/guidance-ai/llguidance/issues/211)
`try_consume_tokens` short-circuits when it finds an invalid token, missing the `inner.parser.check_stop` check and failing to put the parser into a stopped state. Subsequent calls to `try_consume_tokens` will always fail, and the parser is prevented from ever moving into a stopped state.

The fix:
Call `inner.parser.check_stop` in the short-circuit condition


mmoskal (approved):
Good catch!

PR #213, opened 2025-07-23
[Bug] calling `try_consume_tokens` after parser is stopped causes error state

### LINKED ISSUE #211: request: Restore the `LLMatcher` when it consumes an unacceptable token

hudson-ai:
Avoids `is_initialized` check (which causes an error state if the parser is stopped) in `validate_token` and `validate_tokens_raw` by checking whether the parser is stopped and returning early if so

Fixes #211

### DIFF
--- a/parser/src/tokenparser.rs
+++ b/parser/src/tokenparser.rs
@@ -333,13 +333,6 @@ impl TokenParser {
 
     fn check_initialized(&self, lbl: &str) -> Result<()> {
         ensure!(!self.is_fresh, "process_prompt() not called in {}", lbl);
-        ensure!(
-            self.stop_reason == StopReason::NotStopped,
-            "parser stopped in {}; {}",
-            lbl,
-            self.error_message()
-                .unwrap_or("no error message".to_string())
-        );
         Ok(())
     }
 
--- a/parser/src/tokenparser.rs
+++ b/parser/src/tokenparser.rs
@@ -139,6 +139,10 @@ impl TokenParser {
         self.stop_reason
     }
 
+    pub fn stopped(&self) -> bool {
+        self.stop_reason != StopReason::NotStopped
+    }
+
     pub fn is_fresh(&self) -> bool {
         self.is_fresh
     }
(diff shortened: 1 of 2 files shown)


mmoskal:
I think is would be safer to do `if self.stopped() { return false }` (or 0) in validate_token and validate_tokens_raw

hudson-ai:
> I think is would be safer to do `if self.stopped() { return false }` (or 0) in validate_token and validate_tokens_raw

Super reasonable :)

mmoskal (approved):

PR #214, opened 2025-07-23
[Test] Further bounds testing

riedgar-ms:
More detailed tests of the bounds on `integer` and `number` types. As part of this, the `json_schema_check()` routine has been reworked to match the requirements of the JSON testing case (basically, each incoming test `Value` is complete, so if the grammar hasn't matched by the end, there's a problem).

The `integer` bounds tests all work. Unfortunately, as detailed in #216 , there do appear to be residual problems in the `number` bounds. For this reason, I have not expanded the more comprehensive test approach with a `number_both_bounds()` test; there is little point until the individual bounds tests work in full.


sample_parser/tests/common_lark_utils/mod.rs
```diff
@@ -126,9 +126,53 @@ pub fn lark_str_test_many_ext(quiet: bool, lark: &str, passing: &[&str], failing
 }
 
 pub fn json_schema_check(schema: &Value, json_obj: &Value, expect_valid: bool) {
```
hudson-ai, line 128: The rewrite was the way to go here :)

PR #229, opened 2025-08-11
[BUG] `force_lexeme_end` -> `try_lexeme_end` in lark lexer when out of input

### LINKED ISSUE #228: Malformed Lark causes panic/backtrace
A malformed Lark with an unclosed terminal causes a panic/backtrace, e.g. `start: / "test"\n` when run through `LLMatcher.validate_grammar(llg_grammar, limits=LLParserLimits())`.

Specific error:
```
thread '<unnamed>' panicked at parser/src/lark/parser.rs:543:30:
begin <= end (1 <= 0) when slicing `/ "test"
```

hudson-ai:
Calls `try_lexeme_end` instead of `force_lexeme_end` when we run out of input while lexing lark, avoiding a click of the ruby slippers.

@mmoskal I'd appreciate your eyes on this! My inability to get the string case to fail pre-"fix" does make me feel a little un-confident that I understood the problem correctly.

Fixes #228

### DIFF
--- a/parser/src/lark/lexer.rs
+++ b/parser/src/lark/lexer.rs
@@ -279,7 +279,7 @@ pub fn lex_lark(input: &str) -> Result<Vec<Lexeme>> {
     while idx <= input_bytes.len() {
         let mut b = b'\n';
         let res = if idx == input_bytes.len() {
-            lexer.force_lexeme_end(state)
+            lexer.try_lexeme_end(state)
         } else {
             b = input_bytes[idx];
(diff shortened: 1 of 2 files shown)


mmoskal (approved):
It crashes all right for me before the fix (when I run your new test), similar to [posted crash](https://community.openai.com/t/gpt-5-custom-lark-tool-outputs-are-not-guaranteed-to-conform-to-the-cfg/1337673/16)

Thank you!

PR #230, opened 2025-08-12
[Test] More JSON testing

riedgar-ms:
Extend the testing of JSON in the Rust layer:
- Set operations on subschema via `anyOf`, `allOf` and `oneOf` (although the last appears to need particular features enabled?)

### DIFF
--- a/sample_parser/tests/common_lark_utils/mod.rs
+++ b/sample_parser/tests/common_lark_utils/mod.rs
@@ -13,6 +13,12 @@ use llguidance::{
 use sample_parser::*;
 use serde_json::Value;
 
+#[derive(Debug)]
+pub enum NumericBounds {
+    Inclusive,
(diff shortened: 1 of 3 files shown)


hudson-ai:
Re: `oneOf` -- 
It is essentially an XOR, and we can't actually support the *exclusive* part. For some sets of schemas, we can determine that all of the options are mutually exclusive, i.e. XOR is equivalent to OR. Otherwise, we require a flag to be explicitly set that says "I'm okay approximating XOR as OR, and I accept the risk that comes with such an approximation".

riedgar-ms:
Makes sense - some of the `oneOf` tests are going through, but there was one where I had to convert to `anyOf` (not sure why it couldn't figure out the schema compatibility, though).

Anyhow, this PR now bumps up the coverage by a few more lines

PR #231, opened 2025-08-12
[Build] Run code coverage from Python

riedgar-ms:
Extending the code coverage workflow with a separate job running the Python tests


riedgar-ms:
The overall coverage from this is surprisingly low; I'm not sure why.

PR #235, opened 2025-08-21
Add option for initial skip of `%ignore` lexeme

### LINKED ISSUE #233: Ignored terminal is reject on the first token
I am trying to use LLGuidance to generate Java program, but I found LLGuidance might reject the first token mysteriously. After debugging, I get the following reproduction code.

In this code, we only accept 'whitespaces', and C-style block comments are ignored. When the first character is part of the ignored block comment, LLGuidance rejects it. If a space is added before the first token, LLGuidance can accept the text normally.

I suspect this issue is related to #123.

I am using LLGuidance 1.2.0 and Qwen 2.5.

```Python
import transformers
import llguidance
import llguidance.hf
import llguidance.torch
import torch

grammar = r'''%llguidance { "no_forcing": true }
start: WS*
%import common.WS
IGNORED: /\/\*.*?\*\//s
%ignore IGNORED
'''

tokenizer = transformers.AutoTokenizer.from_pretrained('Qwen/Qwen2.5-0.5B')
llg_tokenizer = llguidance.hf.from_tokenizer(tokenizer)
mask_data = llguidance.torch.allocate_token_bitmask(1, llg_tokenizer.vocab_size)
interp = llguidance.LLMatcher(llg_tokenizer, grammar)
assert not interp.is_error(), interp.get_error()

def process_logit(input_ids: list[int], logit: torch.Tensor) -> torch.Tensor:
  if input_ids:
    last_token = input_ids[-1]
    assert mask_data[0, last_token // 32] & (1 << (last_token % 32)) != 0
    interp.consume_token(last_token)
    assert not interp.is_error(), interp.get_error()
  llguidance.torch.fill_next_token_bitmask(interp, mask_data)
  llguidance.torch.apply_token_bitmask_inplace(logit, mask_data)
  return logit

def process_tokens(tokens: list[int]):
  for i in range(len(tokens)):
    logit = torch.zeros((llg_tokenizer.vocab_size,), dtype=torch.float)
    visible_tokens = tokens[:i]
    masked_logit = process_logit(visible_tokens, logit)
    assert masked_logit[tokens[i]] != float('-inf'), f'token {i} ({tokens[i]}, {tokenizer.decode(tokens[i])!r}) is masked'

text = '/*  */'
tokens = tokenizer.encode(text)
print(tokens)
process_tokens(tokens)
```

hudson-ai:
Resolves #233 by exposing an `allow_initial_skip` option

### DIFF
--- a/parser/src/api.rs
+++ b/parser/src/api.rs
@@ -47,6 +47,11 @@ pub struct LLGuidanceOptions {
     /// This is very unlikely what you need.
     #[serde(default)]
     pub allow_invalid_utf8: bool,
+
+    /// If set, the grammar will allow the %ignore lexeme at the start of the grammar.
+    /// Otherwise, it will only be allowed after the first non-ignored lexeme.
+    #[serde(default)]
+    pub allow_initial_skip: bool,
 }
 
 impl LLGuidanceOptions {
@@ -56,6 +61,9 @@ impl LLGuidanceOptions {
         }
         if other.allow_invalid_utf8 {
             self.allow_invalid_utf8 = true;
+        }
+        if other.allow_initial_skip {
+            self.allow_initial_skip = true;
         }
     }
 }
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -318,7 +318,7 @@ MULT_NUM: %regex {
 
 Certain grammar options can be set by using `%llguidnace { ... }`,
 by passing it a JSON object with the options;
-see `LLGuidanceOptions` in [api.rs](../parser/src/api.rs#L24).
+see `LLGuidanceOptions` in [api.rs](../parser/src/api.rs#L36).
 Example: `%llguidance { "no_forcing": true }`.
 It can be specified multiple times, with the options being merged.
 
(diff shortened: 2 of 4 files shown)


parser/src/api.rs
```diff
@@ -47,6 +47,11 @@ pub struct LLGuidanceOptions {
     /// This is very unlikely what you need.
     #[serde(default)]
     pub allow_invalid_utf8: bool,
+
+    /// If set, the grammar will allow the %ignore lexeme at the start of the grammar.
+    /// Otherwise, it will only be allowed after the first non-ignored lexeme.
```
mmoskal, line 52: Note that, like the other options here, this applies to the entire grammar, with all sub-grammars.

Maybe worth a comment? (the other options are really not to be used, but this one may be)

mmoskal (approved):

parser/src/api.rs
hudson-ai: Agreed -- added the comment. My apologies for leaving this to sit for so long!

PR #243, opened 2025-09-26
Bump MSRV to 1.87.0

hudson-ai:
Bump MSRV to `1.87` and fix clippy lints.

@mmoskal ping in case you oppose bumping MSRV -- otherwise we can pin the version used for linting checks (in order to unblock PRs).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -62,7 +62,7 @@ jobs:
     - uses: actions/checkout@v4
 
     - name: Set up Rust
-      uses: dtolnay/rust-toolchain@1.90.0
+      uses: dtolnay/rust-toolchain@1.87.0
       with:
         components: clippy
 


python_ext/src/py.rs
```diff
@@ -30,13 +29,6 @@ pub(crate) struct LLTokenizer {
     factory: Arc<ParserFactory>,
 }
 
-#[derive(Serialize, Deserialize)]
-struct PyMidProcessResult {
-    progress: Vec<ParserOutput>,
-    stop: bool,
-    temperature: f32,
-}
-
```
hudson-ai, line 39: Feels rather bizarre that this dead code was not detected before...

mmoskal:
My understanding is that the GH action uses the latest stable rust for clippy, and only uses 1.80 to check if it builds (without clippy). Do your changes break build with 1.80?

Now that llg is so widely used I would be weary or setting MSRV to the latest.

mmoskal:
Oh, I guess we should set `rust-version = something` in cargo.toml so that clippy won't complain about features not supported. We could go for say 1.86 (6 months ago) for both rust-version and the MSRV Check action. I guess we could even for 1.89, but 1.90 seems a bit excessive (a few days ago).

hudson-ai:
> My understanding is that the GH action uses the latest stable rust for clippy, and only uses 1.80 to check if it builds (without clippy). Do your changes break build with 1.80?

Sadly these changes are necessary to make clippy happy when using latest stable (1.90.0). But these changes do indeed break build on 1.80.0. Namely the use of `is_multiple_of`.

> Now that llg is so widely used I would be weary or setting MSRV to the latest.

More or less agreed, although as long as it comes with a (minor at least) bump, I'm not sure it's such a huge problem. That being said, pinning rust to 1.80.0 for clippy checks is a valid (if not conservative) solution.

hudson-ai:
> Oh, I guess we should set `rust-version = something` in cargo.toml so that clippy won't complain about features not supported. We could go for say 1.86 (6 months ago) for both rust-version and the MSRV Check action. I guess we could even for 1.89, but 1.90 seems a bit excessive (a few days ago).

1.86 seems like a nice compromise :)

mmoskal:
Maybe 1.87 so you get the is_multiple_of ;)

PR #244, opened 2025-10-13
[BUG] Inconsistent handling of the EOS token in `try_consume_tokens`

### LINKED ISSUE #239: Inconsistent handling of the EOS token
When I give a list of tokens, ending with EOS token, to try_consume_tokens, then it accepts all of them. But when I give the tokens one after another, then the EOS token is rejected. Is this an expected behaviour? I would rather expect that no matter how I split a list of tokens, the behaviour should be the same.

```
import llguidance, llguidance.hf, transformers
grammar_str = 'root ::= "a"'
model_id = "hsultanbey/codegen350multi_finetuned"
tokenizer = transformers.AutoTokenizer.from_pretrained(model_id)
ll_grammar = llguidance.grammar_from("grammar", grammar_str)
ll_tokenizer = llguidance.hf.from_tokenizer(tokenizer)
ll_matcher = llguidance.LLMatcher(ll_tokenizer, ll_grammar)
a = tokenizer("a")['input_ids'][0]
eos = tokenizer.eos_token_id

print(ll_matcher.try_consume_tokens([a, eos]))  # returns 2
ll_matcher.reset()
print(ll_matcher.try_consume_tokens([a]))  # returns 1
print(ll_matcher.try_consume_tokens([eos]))  # returns 0, but I'd expect 1
```

hudson-ai:
Previously, an EOS token would be allowed to be consumed if `is_accepting && !can_advance` if and only if it was included in the token slice that advanced the parser to this state (being rejected if passed later). 

This PR calls `check_stop` is called after EACH token is consumed, rather than just once before returning from `try_consume_tokens`, ensuring that `try_consume_tokens(X) + try_consume_tokens(Y) = try_consume_tokens(X + Y)` for all token slices `X` and `Y`. 

Fixes #239

### DIFF
--- a/sample_parser/tests/test_raw_parser.rs
+++ b/sample_parser/tests/test_raw_parser.rs
@@ -329,3 +329,29 @@ fn test_try_consume_after_stop() {
     }
     unreachable!();
 }
+
+#[test]
+/// Test that try_consume_tokens accepts a consistent number of tokens whether
+/// 1. the EOS token is included in the input
+/// 2. the EOS token is not included, but is consumed separately afterwards
+///
+/// Note that this test is not opinionated about whether the EOS token is accepted
+/// in either case, just that the total number of tokens consumed is the same.
+fn test_try_consume_eos_consistency() {
+    let lark = r#"start: "a""#;
+    let parser = make_parser(lark);
+    let tokens = get_tok_env().tokenize("a");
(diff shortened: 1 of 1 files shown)


riedgar-ms:
I'm OK with this, so long as we have that test to 'pin' the behaviour. Whether this is what should be happening is a @mmoskal  question :-)

Issue #240, opened 2025-09-18
Bug: Quantified whitespace patterns in x-guidance ignore bounds and accept unlimited whitespace


rajbirs-cerebras:
Hello LLGuidance Team, I am facing an issue with `whitespace_pattern` field in `x-guidance`. [Custom whitespace patterns](https://github.com/guidance-ai/llguidance/blob/94fa39128ef184ffeda33845f6d333f332a34b4d/docs/json_schema.md#whitespace-handling) with quantifiers (e.g., {1,2}) in LLGuidance JSON schema compilation do not enforce their specified bounds and accept unlimited whitespace characters.

### REPRODUCTION
**LLGuidance Version :** v1.2.0
**Schema:**

```
{
  "x-guidance": {
    "whitespace_pattern": r"[\x20\x0A\x0D\x09]{1,2}",
    "whitespace_flexible": false
  },
  "type": "object", 
  "properties": {
    "message": {"type": "string"}
  },
  "required": ["message"],
  "additionalProperties": false
}
```

**Expected Behavior:** Pattern {1,2} should accept 1-2 whitespace characters and reject 0 or 3+ characters.

**Actual Behavior:** 
**Incorrectly Accepts:** `{   "message"   :   "Hello"   }` (3 spaces - should be rejected)
**Correctly Accepts:** `{ "message" : "Hello" }` (1 space)

I also tried to enforce this pattern directly in LLGuidance [code](https://github.com/guidance-ai/llguidance/blob/94fa39128ef184ffeda33845f6d333f332a34b4d/parser/src/json/compiler.rs#L147), but it still does not work and accepts unlimited whitespaces.

Can you please look into the issue and confirm if it's a valid issue or there is a problem with my JSON Schema request? If it's a real issue in LLGuidance, please provide any estimated timeline for the fix.

rajbirs-cerebras:
@mmoskal / @hudson-ai can anyone look into this issue please?

hudson-ai:
Hi @rajbirs-cerebras apologies for the delayed response.

This behavior is "as designed", as the whitespace pattern is treated as an [`%ignore` directive](https://lark-parser.readthedocs.io/en/latest/grammar.html#ignore), which has the semantics of "ignore ALL occurrences of this lexeme" (no matter how many times it is repeated).

[...]

PR #246, opened 2025-10-16
[Feature] Regex JSON separators

hudson-ai:
Allows `item_separator` and `key_separator` (passed in `x-guidance`) to be specified as regex patterns, e.g.

```json
{
   "x-guidance": {
        "item_separator": "\\s{0,2},\\s{0,2}",
        "key_separator": "\\s{0,2}:\\s{0,2}",
        "whitespace_flexible": false,
   }
}
```

Addresses #240

Note that this is a *change in behavior*, as these values were previously interpeted as string literals. I expect that this will cause few problems, however.

### DIFF
--- a/sample_parser/tests/test_json_x_guidance.rs
+++ b/sample_parser/tests/test_json_x_guidance.rs
@@ -1,9 +1,19 @@
 mod common_lark_utils;
 use common_lark_utils::lark_str_test;
+use rstest::rstest;
 use serde_json::json;
 
-#[test]
-fn test_simple_separators() {
+#[rstest]
+// Ok
+#[case::with_spaces(r#"{"a": 1, "b": 2}"#, true)]
+// Bad
+#[case::no_spaces(r#"{"a":1,"b":2}"#, false)]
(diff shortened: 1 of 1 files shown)


mmoskal:
Maybe better to bite the bullet and change the semantics of these two fields to be regexes? I highly doubt anyone is using them with regex meta-characters. Otherwise, I'm cool with `{"pattern": ...}`.

hudson-ai:
@mmoskal you are probably right. Would certainly make the python integration easier, and I agree that it's probably not a breaking change for a vast, vast majority of use cases. @riedgar-ms would this semantic change be agreeable to you (re-interpreting separators as regexes rather than literal strings)?

riedgar-ms:
I wouldn't object - I agree that it's unlikely that anyone's changing their JSON to be something like `{ "a" ? 1 }`

rajbirs-cerebras:
Test the PR locally and it works fine. Thanks @hudson-ai for preparing this change.

hudson-ai:
@mmoskal will merge today and do a minor bump if you don't object :)

mmoskal:
@hudson-ai sounds good!

PR #249, opened 2025-11-06
fix: tokenization of special tokens for ByteTokenizer

daulet:
Otherwise `tokenize_special` does not respect special tokens present in config.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -5,6 +5,16 @@ on:
     branches: [ "main" ]
   pull_request:
     branches: [ "main" ]
+  workflow_dispatch:
+    inputs:
+      commit_id:
+        description: 'Branch or Commit ID (optional)'
+        required: false
+        type: string
+  schedule:
+    # * is a special character in YAML so we quote this string
+    # Run at 09:10 UTC every day
+    - cron:  '10 09 * * *'
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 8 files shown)


hudson-ai:
Thanks for the PR @daulet!

I think I'd feel more comfortable adding something like 
```rust
    fn tokenize_bytes_special(&self, s: &[u8]) -> Vec<TokenId> {
        self.tok_trie.tokenize_with_greedy_fallback(s, |s| {
            self.tok_trie.tokenize_with_special(s, |s| {
                self.tokenizer
                    .hf_tokenizer
                    .encode(s, false)
                    .expect("tokenizer error")
                    .get_ids()
                    .to_vec()
            })
        })
    }
```
to the `impl TokenizerEnv for ByteTokenizerEnv`, rather than changing the default impl in the trait itself.

@mmoskal any opinion / think I'm missing something?

mmoskal:
@hudson-ai yes, a specific method for ByteEnv would be better

daulet:
> @hudson-ai yes, a specific method for ByteEnv would be better

hm that was my original patch but i noticed that default implementation completely ignores special tokens which seems like a foot gun for other trait implementations

daulet:
@hudson-ai updated, ptal

hudson-ai:
> hm that was my original patch but i noticed that default implementation completely ignores special tokens which seems like a foot gun for other trait implementations

This is a fair point, but I still feel that the "regular" tokenization function (which may very well correctly tokenize special tokens as special) is a safer default impl here, where `self.trie.tokenize_with_greedy_fallback` and `self.trie.tokenize_with_special` are a bit overly opinionated for the generic case (not to mention that they might get called in a "nested" fashion if they are also called in `self.tokenize`).

Anyway, thank you for the fix!

PR #251, opened 2025-11-10
Fix build issues by adding mypy `no_site_packages` flag

hudson-ai:
Add `no_site_packages` flag to mypy config, allow mypy's `python_version` to reflect our minimum supported python version (3.9) while still running tests on a newer python version (3.11).

In particular, pytest (a dev dependency) recently released a version that uses pattern matching (>3.10 syntax). Without this flag, mypy will check pytest and fail.


hudson-ai:
Closing in favor of #252

PR #254, opened 2025-11-17
[Build] Disk space logging

riedgar-ms:
We had some runs fail due to a lack of disk space. Although this appears to have passed, add some extra logging so that we can diagnose issues in future.


hudson-ai (approved):
More logging is good 🫡

riedgar-ms:
> More logging is good 🫡

Well, until the logs eat all the disk space....

PR #260, opened 2025-11-20
Mark immutable pyclasses as frozen

ngoldbaum:
See [the PyO3 docs](https://pyo3.rs/v0.27.1/class.html#frozen-classes-opting-out-of-interior-mutability) for more about the `frozen` option for the pyclass macro.

In short, it tells PyO3 that the Rust data wrapped in the class cannot be mutated via a `&mut` reference. It also makes it a compiler error to create mutable references. It's possible in the future `frozen` will become the default.

The main benefit is performance. Frozen pyclasses do not need to do any atomic reference counting when rust data gets borrowed. It also makes these types a little easier to reason about in a multithreaded context. Since all these types can be trivially marked as `frozen`, the default atomic reference counting scheme in PyO3 is unnecessary.

It's possible the other two pyclasses in the library could be marked as `frozen` as well with a little more work by leveraging some form of interior mutability.

### DIFF
--- a/python_ext/src/llmatcher.rs
+++ b/python_ext/src/llmatcher.rs
@@ -22,7 +22,7 @@ struct LLMatcher {
     tok_env: TokEnv,
 }
 
-#[pyclass]
+#[pyclass(frozen)]
 struct LLExecutor {
     pool: rayon::ThreadPool,
 }
--- a/python_ext/src/parserlimits.rs
+++ b/python_ext/src/parserlimits.rs
@@ -1,7 +1,7 @@
 use llguidance::api::ParserLimits;
 use pyo3::prelude::*;
 
-#[pyclass]
+#[pyclass(frozen)]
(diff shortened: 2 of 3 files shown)


riedgar-ms:
Thank you for contributing this! Does this mean that any attempt from Python to modify the contents of these `structs` will hit a runtime error?

ngoldbaum:
It means it's not currently possible to mutate these from Python. If it were, I couldn't trivially mark them as frozen, because that would conflict with a method that accepts a `&mut self` argument, for example.

It's also possible to add python-level mutability to a frozen pyclass by leveraging interior mutability. One way to think about frozen pyclasses is they turn off the default ARC-based interior mutability for runtime borrow checking and allow users to add their own interior mutability implementation.

riedgar-ms:
Thanks for the clarification!

PR #264, opened 2025-12-16
Speed up toktrie add_bias (~3% improvement)

omarkilani:
## Summary
  - Add Criterion benchmarks for `compute_mask` hot path
  - Optimize `add_bias_inner` trie walk with ~3% improvement

  ## Changes

  ### Benchmarks
  Added two Criterion benchmarks to measure the `compute_mask` hot path:
  - `compute_mask/title_string/32k`: measures mask computation during string value parsing
  - `generate_string/10tok/32k`: measures realistic token generation with rollback

  ### Performance optimizations
  - Use `#[inline(always)]` on `add_bias_inner` for consistent inlining
  - Use `unsafe get_unchecked` for node access in the hot loop
  - Add `allow_token_unchecked()` for bounds-check-free token setting

### DIFF
--- a/toktrie/src/svob.rs
+++ b/toktrie/src/svob.rs
@@ -296,7 +296,7 @@ impl SimpleVob {
     }
 
     pub fn resize(&mut self, size: usize) {
-        let new_size = (size + BITS - 1) / BITS;
+        let new_size = size.div_ceil(BITS);
         assert!(new_size >= self.data.len());
         self.data.resize(new_size, 0);
         self.size = size;


hudson-ai:
Thanks a bunch for the contribution @omarkilani!

omarkilani:
> Thanks a bunch for the contribution @omarkilani!

Thanks for your hard work on llguidance. 🫡

Not sure what the rules are around unsafe for the project in general but thought I'd put this here in case it's helpful.

riedgar-ms:
I'm not familiar enough with the code to have strong opinions about this.... it looks like the new unsafe function isn't _quite_ a drop-in replacement for the previous version?

Does this mean that we should be adding a `cargo bench` workflow?

hudson-ai:
> I'm not familiar enough with the code to have strong opinions about this.... it looks like the new unsafe function isn't _quite_ a drop-in replacement for the previous version?
> 
> Does this mean that we should be adding a `cargo bench` workflow?

No, not a drop-in replacement. The unsafe function is functionally equivalent, however it is faster at the expense of losing compile-time guarantees of memory safety. It's used in this case to speed up the hot loop (`while p < endp`...), and a run-time check is added before the loop to ensure that the right invariant is being enforced (`let nodes = &self.nodes[..endp]`).

Other usage of the "safe" version of this function have not been replaced.

Re: `cargo bench` -- it's honestly not a terrible idea to add some kind of ratchet to the PR gate, but it's certainly outside the scope of this PR. Can take this discussion offline :)

riedgar-ms (approved):
I've no objection to hoisting a bounds check outside a loop. I agree that setting this up in a workflow is beyond the scope of this PR (and a skim of the criterion docs counsels against such a move).

hudson-ai:
Hey @omarkilani just checking back in. I was about to just revert the inlining behavior and merge, but I realized that I made a bit of an oversight error the first time that I reviewed. Just want to give you an opportunity to take a look yourself before I take action.

The use of `defl_tok` was a nice trick that was being used to avoid the conditional `if let Some(tok) = n.token_id()` check in the hot loop. It seems that your unsafe optimization outweighs the cost of removing the trick, but can we not get the best of both worlds, i.e. do
```rust
unsafe { toks.allow_token_unchecked(n.token_id().unwrap_or(defl_tok)) };
```
with no `if let` guard?

It *should* respect the invariant if I'm thinking about this correctly, as `alloc_token_set()` allocates capacity for `vocab_size + 1` tokens.

omarkilani:
Hey @hudson-ai - I'm now at NVIDIA and will need to figure out a bunch of things to push new commits.

I think it would be faster for you to take over if you'd like as the commits pushed while at Groq are okay.

Apologies for the delay in responding.

hudson-ai:
@omarkilani appreciate the quick response, and no worries about the earlier delay! I'm happy to take over from here -- thanks for the consent. Congrats on the new job 😄

hudson-ai:
@riedgar-ms and/or @JC1DA would love some more eyes to make sure I'm not making a mistake on the invariants required by the unsafe code here

testonlyinprod:
@hudson-ai changes are made, if it looks good feel free to merge!

PR #271, opened 2026-01-22
Add parallel token consumption API (consume_token_par)

alecsolder:
## Summary
- Adds `consume_token_par` method to `LLExecutor` for consuming tokens across multiple matchers in
parallel
- Exposes the API through both `llguidance.numpy` and `llguidance.torch` modules
- Complements existing parallel bitmask methods (`fill_next_token_bitmask_par`)

## Test plan
- Added `test_consume_token_par()` verifying parallel consumption with valid, invalid, and mixed token
scenarios
- Added `test_consume_token_par_errors()` testing edge cases (empty list, wrong tuple size, already
borrowed)

### DIFF
--- a/python/llguidance/_lib.pyi
+++ b/python/llguidance/_lib.pyi
@@ -584,6 +584,10 @@ class LLExecutor:
 
         Returns:
             List[bool]: Success/failure for each matcher (in order).
+
+        Note:
+            Matchers that fail (return False) are left in an error state,
+            consistent with the behavior of consume_token on a single matcher.
         """
 
 class JsonCompileOptions(TypedDict, total=False):
--- a/python/llguidance/numpy.py
+++ b/python/llguidance/numpy.py
@@ -93,5 +93,9 @@ def consume_token_par(executor: LLExecutor,
 
     Returns:
         List[bool]: Success/failure for each matcher (in order).
+
+    Note:
+        Matchers that fail (return False) are left in an error state,
+        consistent with the behavior of consume_token on a single matcher.
     """
     return executor.consume_token_par(matchers)
--- a/python/llguidance/torch.py
+++ b/python/llguidance/torch.py
@@ -90,5 +90,9 @@ def consume_token_par(executor: LLExecutor,
 
     Returns:
         List[bool]: Success/failure for each matcher (in order).
+
+    Note:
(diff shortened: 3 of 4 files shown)


hudson-ai:
Hey @alecsolder thanks for the PR! I agree that a parallel `consume_token` api makes total sense and indeed compliments the existing parallel bitmask methods.

What do you think about doing something similar for `consume_tokens` and `try_consume_tokens`? I suppose that adding ALL of these to the API is a bit verbose, but I imagine that `try_consume_tokens` may be the most useful/flexible method if we were only going to expose one. But one of the "plural" ones is likely going to be critical to make parallel consumption useful for use in speculative decoding.

Is there a specific use-case you are hoping to support here?

@JC1DA @riedgar-ms ping for visibility

alecsolder:
> Hey @alecsolder thanks for the PR! I agree that a parallel `consume_token` api makes total sense and indeed compliments the existing parallel bitmask methods.
> 
> What do you think about doing something similar for `consume_tokens` and `try_consume_tokens`? I suppose that adding ALL of these to the API is a bit verbose, but I imagine that `try_consume_tokens` may be the most useful/flexible method if we were only going to expose one. But one of the "plural" ones is likely going to be critical to make parallel consumption useful for use in speculative decoding.
> 
> Is there a specific use-case you are hoping to support here?
> 
> @JC1DA @riedgar-ms ping for visibility
Hey @hudson-ai ,

Hello! I'm specifically looking for a parallel counterpart to fill_next_token_bitmask_par, I don't specifically need speculative decoding support atm, I am just looking to parallelize token acceptance to speed up high batch size constrained sampling workloads in vLLM. Does that make sense? 

For example, we have 300 matchers and 300 tokens to accept, and doing it iteratively ends up being quite slow, so having a parallel rust API would be helpful. If there is a better way to do it here, more than happy to!

hudson-ai:
@alecsolder the `consume_tokens` and `try_consume_tokens` variants can come later if there is demand for the features (from you or others), but this feels like a pretty clean and minimal change that supports your use-case as-is.

Let me take a closer look at the implementation itself

hudson-ai:
Thank you!

alecsolder:
> Thank you!

Perfect, thanks so much!

PR #279, opened 2026-02-03
Mask caching for lazy lexemes

### LINKED ISSUE #275: Lazy Lexeme Performance
Hi there, 

We've noticed that performance significantly degrades inside of a lazy lexeme, and were wondering if there is anything that can be done about it.
For example, a lazy rule where the tool parameter tag is not a single token in the model's vocab. This rule is:
```
end_parameter_head[lazy]: TEXT_CONTENT "</parameter>"
parameter_end_rule: end_parameter_head
TEXT_CONTENT: /.*/s
```
We fed some tokens into LLGuidance with the full grammar and plotted the compute time for each token, shown below.

Image

Across various input samples, we see much slower compute times when we are inside of a lazy rule. 

We compared this to the non-lazy case. We used the same exact input minus replacing `"</parameter>"` with an arbitrary single token.
```
parameter_end_rule: TEXT_CONTENT <[some token id]>
TEXT_CONTENT: /.*/s
```
We do see spikes in latency, which is to be expected, but not a distinct "slowdown" period tied to a specific part of the grammar as is the case for the lazy rule.

Image

Is there any plan to introduce optimizations to lazy rules? Is there an obvious optimization that we can apply to the aforementioned lazy rule that we are perhaps missing?

hudson-ai:
Introduces a cache to avoid re-computing the bias/mask whenever the lexer state is unchanged after pushing a token (which is a common case when inside of a permissive lazy lexeme like `rule[lazy]: /.*/ "END"`). It's essentially implemented as an LRU cache with max size 1.

@mmoskal would really appreciate your eyes on this one. I'm using (`lexer_state`, `row_idx`) as the cache key, which I *believe* is right, but please let me know if there are any cases you can imagine this causing incorrect results for. There are pretty good improvements across the board (which actually makes me a little bit wary about correctness...)

Impacts on maskbench (TTFM changes probably noise, but the TBM changes are quite significant):
| Metric (us)   | OLD      | NEW      | Change |
|-----------|----------|----------|--------|
| TBM avg   | 36       | 29       | -19%   |
| TBM p25   | 12       | 7        | -42%   |
| TBM p50   | 21       | 13       | -38%   |
| TBM p75   | 27       | 24       | -11%   |
| TBM p90   | 48       | 41       | -15%   |
| TBM p95   | 80       | 71       | -11%   |
| TBM p99   | 421      | 327      | -22%   |
| TBM p99.9 | 1,316    | 1,238    | -6%    |
| TBM p100  | 10,982   | 11,221   | +2%    |
| TTFM avg  | 1,458    | 1,422    | -2%    |
| TTFM p25  | 661      | 647      | -2%    |
| TTFM p50  | 873      | 842      | -4%    |
| TTFM p75  | 1,335    | 1,285    | -4%    |
| TTFM p90  | 2,332    | 2,236    | -4%    |
| TTFM p95  | 3,561    | 3,447    | -3%    |
| TTFM p99  | 13,342   | 13,046   | -2%    |
| TTFM p99.9| 51,588   | 52,929   | +3%    |
| TTFM p100 | 134,224  | 132,332  | -1%    |

Aims to fix #275

### DIFF
--- a/parser/src/earley/parser.rs
+++ b/parser/src/earley/parser.rs
@@ -468,8 +468,8 @@ struct ParserState {
     parser_error: Option<String>,
     backtrack_byte_count: usize,
 
-    // Cache for compute_bias - avoids recomputing identical masks
-    // when lexer state hasn't changed (common in lazy lexemes)
+    // Cache for compute_bias - avoids recomputing identical masks when lexer state hasn't changed
+    // (common in long lexemes, e.g. the interior of JSON strings)
     bias_cache: Option<BiasCache>,
 
     shared_box: Box<SharedState>,


hudson-ai:
Note: this optimization works for greedy lexemes too, but the gains are smaller because greedy lexemes are already extremely efficient to compute. Lazy lexemes have slicer optimizations disabled, which is the primary source of their overhead.

mmoskal:
Maybe:
- you need to key also on `has_pending_lexeme_bytes()`
- don't do it if max_tokens occurs in the grammar - I think it will be wrong then (the lexer_state/row_idx will be the same no matter how many tokens you process; though maybe not, it seems we're calling this `limit_state_to()` thingy)
- try to have some option that enables this and run both and compare equality in testsuite

Not very sure about any of this.

hudson-ai:
@mmoskal thank you for the pointers!

mmoskal:
Also OOC are there lazy lexemes in the maskbench?

hudson-ai:
> Also OOC are there lazy lexemes in the maskbench?

No there aren't -- my PR title is a bit of a misnomer. Lazy lexemes are several orders of magnitude slower than their greedy counterparts, so caching yields very very significant gains. Greedy lexemes still see some speedup though, as evidenced by the maskbench results above

JC1DA:
@hudson-ai 
I understand the concept and the code looks good for me but I don't think I have enough understanding of llguidance to say if it's missing something or not.
One suggestion is to add a flag to enable/disable this feature (disabled by default) so if there is any issue with this, it will not cause problems to existing users.

PR #282, opened 2026-02-05
Update PyO3 and add free-threaded CI testing

ngoldbaum:
To enable testing on the free-threaded build, I added a build matrix based on the Python version in the CI.

Everything except for the Python tests runs twice, I could probably refactor a bit more but I wanted to propose something more minimal and get feedback before making a bigger change.

Since `torch.compile` isn't yet supported on the free-threaded build, I had to add skips for tests that depend on it if importing `torch.compile` fails.

I also updated PyO3 to the latest version.

Let me know if you'd prefer I split this up into two PRs. I need to update PyO3 to avoid warnings about re-enabling the GIL. PyO3 0.28 requires explicit opt-out of free-threaded support.

### DIFF
--- a/scripts/test-guidance.sh
+++ b/scripts/test-guidance.sh
@@ -116,6 +116,15 @@ echo "Running mypy..."
 
 PYTEST_FLAGS=
 
+if [ "$(python -c "import sysconfig; print(sysconfig.get_config_var('Py_GIL_DISABLED'))")" -eq "1" ]; then
+    # force-disable the GIL because some dependencies do not yet declare support
+    # TODO: delete when this is no longer the case for tokenizers, safetensors,
+    # and llama-cpp-python
+    PYTHON_FLAGS=-Xgil=0
+else
+    PYTHON_FLAGS=
+fi
+
 if test -f ../guidance/tests/unit/test_ll.py ; then
     echo "Guidance side by side"
     cd ../guidance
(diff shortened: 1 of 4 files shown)


riedgar-ms:
Thank you for looking at this! I will defer to others on the PyO3 changes, but I do know that the workflow needs a bit of refactoring.

ngoldbaum:
> but I do know that the workflow needs a bit of refactoring

OK, I'll look at splitting out either the Python tests or just the free-threaded Python tests into their own job.

riedgar-ms:
> > but I do know that the workflow needs a bit of refactoring
> 
> OK, I'll look at splitting out either the Python tests or just the free-threaded Python tests into their own job.

That wasn't supposed to be a criticism of this PR or directed at you! Getting the builds smoother is on my ToDo list :-) My apologies, I should have said "*our* workflow needs a bit of refactoring"

scripts/test-guidance.sh
```diff
@@ -116,6 +116,15 @@ echo "Running mypy..."
 
 PYTEST_FLAGS=
 
+if [ "$(python -c "import sysconfig; print(sysconfig.get_config_var('Py_GIL_DISABLED'))")" -eq "1" ]; then
+    # force-disable the GIL because some dependencies do not yet declare support
+    # TODO: delete when this is no longer the case for tokenizers, safetensors,
+    # and llama-cpp-python
+    PYTHON_FLAGS=-Xgil=0
```
ngoldbaum: I noticed that the GIL was getting re-enabled at runtime in the tests because of test dependencies. This force-disables the GIL in the tests.

Ideally we wouldn't have to do this. I'm actively working with the safetensors and tokenizers maintainers to add 3.14t support. llama-cpp-python unfortunately appears to be unmaintained right now so I don't think there's anyone working on 3.14 or 3.14t support.

ngoldbaum:
Passing CI run from my fork: https://github.com/ngoldbaum/llguidance/actions/runs/21874017567

hudson-ai:
@riedgar-ms ping :)

PR #283, opened 2026-02-06
[FEAT] Support for 'uri' format in JSON

riedgar-ms:
Add 'uri' format support for JSON schema validation

   - Add RFC 3986 compliant URI regex pattern to formats.rs
   - Support all URI components: scheme, authority, path, query, fragment
   - Handle IPv4, IPv6, and reg-name hosts
   - Add valid URI tests including all RFC 3986 Section 1.1.2 examples
   - Add invalid URI tests for edge cases
   - Update documentation to list uri as supported format
   - Update `expected_maskbench.json` with the new results

Some test cases taken from:
https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests/draft2020-12/optional/format/uri.json

### DIFF
--- a/json_stats/expected_maskbench.json
+++ b/json_stats/expected_maskbench.json
@@ -8235,9 +8235,7 @@
   "Github_ultra---o10420.json": {},
   "Github_ultra---o12343.json": {},
   "Github_ultra---o12344.json": {},
-  "Github_ultra---o12652.json": {
-    "validation_error": "test #0: token not accepted at ⟦{\"⟧ * ⟦$⟧ * ⟦schema‧\":\"‧http‧://⟧ mask: TokenSet: 20/128256; ⟨a⟩ ⟨b⟩ ⟨c⟩ ⟨d⟩ ⟨con⟩ ⟨aw⟩ ⟨cont⟩ ⟨co⟩ ⟨date⟩ ⟨da⟩ ⟨dat⟩ ⟨contr⟩ ⟨awa⟩ ⟨bu⟩ ⟨buy⟩ ⟨contract⟩ ⟨buyer⟩ ⟨contracts⟩ ⟨award⟩ ⟨contra⟩"
-  },
+  "Github_ultra---o12652.json": {},
   "Github_ultra---o12653.json": {},
   "Github_ultra---o13020.json": {
     "json_error": "Unimplemented keys: [\"not\"]"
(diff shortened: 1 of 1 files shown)


hudson-ai:
@riedgar-ms some more tests for you!

https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests/draft2020-12/optional/format/uri.json

It would also be good to re-run the JSON Schema Test suite and update any coverage numbers that this PR may affect :)

riedgar-ms:
> @riedgar-ms some more tests for you!
> 
> https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests/draft2020-12/optional/format/uri.json
> 
> It would also be good to re-run the JSON Schema Test suite and update any coverage numbers that this PR may affect :)

A selection should have been added from there

riedgar-ms:
@hudson-ai do you have any thoughts on the maskbench updates? Most are fine, either becoming passes, or hitting the 'oneOf not supported' error. There are a handful which surprise me (although I didn't spot any new failures, just changes in the failure itself), and I'm not sure if those are a concern.

hudson-ai:
@riedgar-ms I'd like to take a closer look at the changes... Will try and make some time today and/or tomorrow to do so!

json_stats/expected_maskbench.json
```diff
@@ -3286,19 +3226,15 @@
     "json_error": "Unimplemented keys: [\"not\"]"
   },
   "Github_hard---o10421.json": {
-    "json_error": "Unknown format: uri"
+    "json_error": "Unsatisfiable schema: incompatible types\n  while processing json-schema:///#/definitions/ContentTypeType"
```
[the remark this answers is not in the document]
riedgar-ms: I suppose the question is the provenance of the original schema. This is implying that the schema itself is bad and shouldn't be working. That's not quite the same as "now we hit the `oneOf` issue" which happens in quite a few cases as well (and which was unconcerning since I knew about that)

json_stats/expected_maskbench.json
```diff
@@ -3337,7 +3273,7 @@
   },
   "Github_hard---o12477.json": {},
   "Github_hard---o12498.json": {
-    "json_error": "Unknown format: uri"
+    "validation_error": "test #0: token not accepted at ⟦{\"‧$‧schema‧\":\"‧http‧://‧json‧-schema‧.org‧/d‧raft‧-‧07‧/schema‧#‧\",\"‧name⟧ * ⟦\":\"⟧ * ⟦my‧-package‧\",\"‧slug⟧ mask: TokenSet: 123231/128256; ALL EXCEPT EOS ⟨\"⟩ ≺HEX[a1]≻ ≺HEX[a2]≻ ≺HEX[a3]≻ ≺HEX[a4]≻ ≺HEX[a5]≻ ≺HEX[a6]≻ ≺HEX[a7]≻ ≺HEX[a8]≻ ≺HEX[a9]≻ ≺HEX[aa]≻ ≺HEX[ab]≻ ≺HEX[ac]≻ ≺HEX[ae]≻ ≺HEX[af]≻ ≺HEX[b0]≻ ≺HEX[b1]≻ ≺HEX[b2]≻ ≺HEX[b3]≻ ≺HEX[b4]≻ ≺HEX[b5]≻ ≺HEX[b6]≻ ≺HEX[b7]≻ ≺HEX[b8]≻ ≺HEX[b9]≻ ≺HEX[ba]≻ ≺HEX[bb]≻ ≺HEX[bc]≻ ≺HEX[bd]≻ ≺HEX[be]≻ ≺HEX[bf]≻ ≺HEX[c0]≻ ≺HEX[c1]≻ ≺HEX[f5]≻ ≺HEX[f6]≻ ≺HEX[f7]≻ ≺HEX[f8]≻ ≺HEX[f9]≻ ≺HEX[fa]≻ ≺HEX[fb]≻ ≺HEX[fc]≻ ≺HEX[fd]≻ ≺HEX[fe]≻ ≺HEX[ff]≻ ⟨\\0⟩ ⟨\\u{1}⟩ ⟨\\u{2}⟩ ⟨\\u{3}⟩ ⟨\\u{4}⟩ ..."
```
riedgar-ms: So I'm trying to understand this error. The relevant file in JSON schema bench:
https://github.com/guidance-ai/jsonschemabench/blob/main/maskbench/data/Github_hard---o12498.json

I've copy/pasted the schema and its example into https://www.jsonschemavalidator.net/ and that says that the schema is OK.

AFAICT, this error is probably the parser objecting to the `"slug"` key in the sample JSON. That is defined [here](https://github.com/guidance-ai/jsonschemabench/blob/07c666b1a2b33e2bcee714e85a8e0ad511823f20/maskbench/data/Github_hard---o12498.json#L103) in the above file, and is listed after the key `"name"`.... so that doesn't look like a field order issue?

I have tweaked the value of `slug` so that it no longer matches the pattern, and then `json_stats` itself complains about not matching the given pattern.

I have also experimented with putting the `slug` key before the `name` key in the data file. In that case, it gives almost the same error:
```
VALIDATION: test #0: token not accepted at ⟦{"‧$‧schema‧":"‧NA‧","‧slug⟧ * ⟦":"⟧ * ⟦my‧-package‧","‧name⟧ mask: TokenSet: 123231/128256; ALL EXCEPT EOS ⟨"⟩ ≺HEX[a1]≻ ≺HEX[a2]≻ ≺HEX[a3]≻ ≺HEX[a4]≻ ≺HEX[a5]≻ ≺HEX[a6]≻ ≺HEX[a7]≻ ≺HEX[a8]≻ ≺HEX[a9]...
```
the main difference here is that the key which appears to be triggering the error is now `name`.

sample_parser/tests/test_json_objects.rs
```diff
@@ -201,3 +201,14 @@ fn linked_list(#[case] obj: &Value) {
 fn linked_list_failures(#[case] obj: &Value) {
     json_schema_check(&LINKED_LIST, obj, false);
 }
+
+#[rstest]
+#[case::neither_required(&json!({"$schema": "http://json-schema.org/draft-07/schema#","a_string": "Hello"}))]
```
riedgar-ms: I think that these tests should pass, but:

https://github.com/guidance-ai/llguidance/actions/runs/21924263215/job/63312527723#step:7:2082

riedgar-ms:
This needs the PR on `jsonschemabench` to merge before the build will clear.

hudson-ai (approved):

PR #287, opened 2026-02-17
Make JSON escape policy configurable via JsonCompileOptions

ostepanian:
- Problem: some users need to disallow \uXXXX escapes during constrained JSON string generation to avoid undesired escape-sequence outputs.
- Current behavior: allowed JSON string escapes are fixed and include u.
- Change: add json_allowed_escapes: Option<String> to JsonCompileOptions.
- Change: wire json_allowed_escapes into json_quote() so JsonQuoteOptions.allowed_escapes is configurable.
- Change: allow this option to be set via JSON Schema x-guidance (through existing JsonCompileOptions deserialization).
- Default behavior: unchanged, still uses nrbtf\\\"u when no override is provided.
- Opt-in behavior: setting json_allowed_escapes without u disables \uXXXX escapes.
- Backward compatibility: preserved by keeping default behavior unchanged.
- Tests added: json_quote_default_allows_u_escape.
- Tests added: json_quote_uses_override_from_options.
- Tests added: x_guidance_accepts_json_allowed_escapes.
- Validation: cargo test -p llguidance json::compiler -- --nocapture.
- Example:
- "x-guidance": { "json_allowed_escapes": "nrbtf\\\"" }

### DIFF
--- a/python_ext/src/py.rs
+++ b/python_ext/src/py.rs
@@ -336,17 +336,19 @@ struct JsonCompiler {
     whitespace_flexible: bool,
     whitespace_pattern: Option<String>,
     coerce_one_of: bool,
+    json_allowed_escapes: Option<String>,
 }
 
 #[pymethods]
 impl JsonCompiler {
     #[new]
-    #[pyo3(signature = (separators = None, whitespace_flexible = false, coerce_one_of = false, whitespace_pattern = None))]
+    #[pyo3(signature = (separators = None, whitespace_flexible = false, coerce_one_of = false, whitespace_pattern = None, json_allowed_escapes = None))]
     fn py_new(
         separators: Option<(String, String)>,
         whitespace_flexible: bool,
         coerce_one_of: bool,
         whitespace_pattern: Option<String>,
+        json_allowed_escapes: Option<String>,
     ) -> Self {
         let (item_separator, key_separator) = separators.unwrap_or_else(|| {
             if whitespace_flexible {
@@ -361,6 +363,7 @@ impl JsonCompiler {
             whitespace_flexible,
             coerce_one_of,
             whitespace_pattern,
+            json_allowed_escapes,
         }
     }
     #[pyo3(signature = (schema, check = true))]
@@ -373,6 +376,7 @@ impl JsonCompiler {
             coerce_one_of: self.coerce_one_of,
             whitespace_pattern: self.whitespace_pattern.clone(),
             lenient: false,
+            json_allowed_escapes: self.json_allowed_escapes.clone(),
             retriever: None,
         };
         compile_options.apply_to(&mut schema);


hudson-ai:
Hey @ostepanian thanks for the contribution!

I took a look, and I have some concerns, specifically regarding conformity to [ECMA 404](https://ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) / [RFC 8259](https://www.rfc-editor.org/rfc/rfc8259). By these standards, any hexadecimal escape sequence should be allowed, even those representing control escapes. E.g. in python, the following works fine: `json.loads(r'"Guillain-Barr\u0000e9 syndrome"')`. 

Note that changes like this likely need to be done in the upstream `derivre` crate, where a bit of [extra work](https://github.com/guidance-ai/derivre/blob/c68f60791d191a05131796771f38c2e904417074/src/regexbuilder.rs#L337) is being done to convert regexes like `/.*/` into regexes that only match valid JSON strings. This is necessary to support cases where users pass their own regexes in the `pattern` field of a JSON schema.

This is fairly subtle stuff and I *do* need to take a closer look. Are you able to provide a bit more context on the actual downstream issue you're hitting?

Unrelated to the PR itself -- @riedgar-ms, we should probably rewrite CI a little bit to make sure that lints / formatting are their own checks so that they don't fail the entire pipeline.

hudson-ai:
Looks like you edited your comment. Will re-read when I get a chance

ostepanian:
> Looks like you edited your comment. Will re-read when I get a chance

Thanks @hudson-ai - yes, we have changed the approach to allow configuring escape sequences instead of hardcoding regexes that are needed for specific cases.

parser/src/json/compiler.rs
```diff
@@ -28,6 +28,10 @@ pub struct JsonCompileOptions {
     pub whitespace_pattern: Option<String>,
     pub coerce_one_of: bool,
     pub lenient: bool,
+    /// Allowed escape letters after '\' when quoting JSON strings.
+    /// Defaults to full JSON set: nrbtf\"u
+    /// For example, set to nrbtf\" to disallow \uXXXX escapes.
```
mmoskal:
```suggestion
    /// Defaults to full JSON set: nrbtf"u\
    /// For example, set to nrbtf"\ to disallow \uXXXX escapes.
```

put \ at the end so it's clear it's not quoted quote

ostepanian:
> please update the docs for x-guidance

done

ostepanian:
> Could you add json_allowed_escapes to the x-guidance docs in docs/json_schema.md? It should go with the other bullet points around line 85. Something like:
> 
> * `json_allowed_escapes`, optional string, defaults to `nrbtf"u\`; controls which escape sequences are allowed after `\` in JSON strings. Each character in the string enables the corresponding escape (e.g., `n` enables `\n`, `u` enables `\uXXXX`). Set to `nrbtf"\` to disallow `\uXXXX` escapes. Note that removing `u` means control
>   characters without named escapes (e.g., U+0000, U+001E) become unrepresentable.

Thanks @hudson-ai - done!

mmoskal (approved):
LGTM, leaving final approval to @hudson-ai

ostepanian:
@hudson-ai - things look green, will not merge until you ack. Thank you!

hudson-ai (approved):
Looks great. Thanks @ostepanian !

PR #299, opened 2026-03-01
Fix decoding of single 0xFF byte tokens

### LINKED ISSUE #248: Fix decoding 0xFF tokens
Some tokens might legitimately be the single 0xFF byte. This can be determined by checking if the length of the token is equal to 1.

I have found that tokenizing "\xFF" gives the correct token id. While decoding that token id in a round trip test simply gives back an empty string.

Fixing the decoding step seems rather trivial by changing this [line](https://github.com/guidance-ai/llguidance/blob/29e33a219f77be1cc37d8394630440f448d19945/toktrie/src/toktree.rs#L412) to check that `tok.len() > 1` like so
```
else if t[0] == TokTrie::SPECIAL_TOKEN_MARKER && t.len() > 1
```

This makes my round trip tests pass.

Note: Ordering the special token check before the length check to short circuit the branch statements more consistently. The length check is already covered by the is_empty() case a few lines earlier.

koriyoshi2041:
## Summary
- A token consisting of a single `0xFF` byte was incorrectly treated as a special token because the check only verified `bytes[0] == SPECIAL_TOKEN_MARKER` without confirming `len() > 1`
- Add a `len() > 1` guard to all 5 affected sites, consistent with the existing correct check in `token_dbg()`
- Affected functions: `decode_ext`, `token_len`, `decode_raw`, `is_special_token` (in `toktree.rs`) and `commit_token_u8` (in `stop_controller.rs`)

Fixes #248

## Test plan
- [ ] `cargo test` passes (no regression)
- [ ] Tokens with a single `0xFF` byte are decoded as the literal byte
- [ ] Multi-byte special tokens (e.g. `0xFF` + name) still work correctly

### DIFF
--- a/parser/src/stop_controller.rs
+++ b/parser/src/stop_controller.rs
@@ -96,7 +96,7 @@ impl StopController {
             self.is_stopped = true;
         } else {
             let bytes = self.tok_env.tok_trie().token(tok_id);
-            if !bytes.is_empty() && bytes[0] == TokTrie::SPECIAL_TOKEN_MARKER {
+            if bytes.len() > 1 && bytes[0] == TokTrie::SPECIAL_TOKEN_MARKER {
                 if let Some(rx) = self.regex.as_mut() {
                     rx.state = rx.initial_state;
                 }
--- a/toktrie/src/toktree.rs
+++ b/toktrie/src/toktree.rs
@@ -483,7 +483,7 @@ impl TokTrie {
 
     pub fn token_len(&self, idx: u32) -> usize {
         let t = self.token(idx);
-        if t.is_empty() || t[0] == TokTrie::SPECIAL_TOKEN_MARKER {
+        if t.is_empty() || (t.len() > 1 && t[0] == TokTrie::SPECIAL_TOKEN_MARKER) {
             let mut idx = idx;
             let mut len = 1;
             while idx >= 10 {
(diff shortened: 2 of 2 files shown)


riedgar-ms:
Thank you for this contribution.... could you add a couple of tests to validate the behaviour?

koriyoshi2041:
Thanks for the review, and sorry for the long delay. I've added regression tests in `toktrie/src/toktree.rs` (`mod tests`).

They build a small `TokTrie` containing a token whose bytes are exactly the single marker byte `0xFF` (token id 1), alongside an ordinary token and a genuine special token (`\xff[7]`), and assert:

- `is_special_token(1)` is `false` (single `0xFF` is a regular byte token), while `is_special_token(2)` stays `true`;
- `token_len(1) == 1` and `token(1) == [0xFF]`;
- `decode_raw([1]) == [0xFF]`.

Each assertion fails on the pre-fix code — e.g. `token_len` returns `4` (the synthetic `\xff[1]` length) and `decode_raw` emits the special-token reference instead of the literal byte — and passes with the `len() > 1` guard. `cargo test -p toktrie`, `cargo fmt --check` and `cargo clippy --tests` are all green.

koriyoshi2041:
Rebased this onto latest main to resolve the conflict in `toktrie/src/toktree.rs`; the conflict was just between the new EOS-token tests upstream and the 0xFF regression tests here, so both sets are kept.\n\nLocal checks after the rebase:\n- `cargo test -p toktrie`\n- `cargo fmt --check`\n- `git diff --check`

PR #320, opened 2026-03-27
[Refactor] Sample Parser & Tests

riedgar-ms:
A lot of the functionality of `parser` was tested in the `sample_parser` crate, which didn't make a lot of sense. Furthermore, the `json_schema_test_suite` took a dependency on several parts of the 'test' portion of `sample_parser`.

This rationalises things. The 'utilitites' portion of `sample_parser` goes into a new workspace-only crate named `llg_test_utils`, while the various parser tests which were also in `sample_parser` move over to `parser/tests`. Both the latter and `json_schema_tests_suite` then take a dependence on `llg_test_utils`


parser/Cargo.toml
```diff
@@ -46,6 +46,12 @@ generate-header = ["dep:cbindgen"]
 [dev-dependencies]
 criterion = "0.8.1"
 regex = "1.11.1"
+toktrie_hf_downloader = { workspace = true }
+llg_test_utils = { workspace = true }
+rstest = "0.25.0"
+serde-json-fmt = "0.1.0"
+lazy_static = "1.5.0"
+serde_json = { version = "1.0.138", features = ["preserve_order"] }
```
hudson-ai, line 54: Just want to be cautious because this is a bit of a blind spot for me... Under what conditions are dev deps pulled in? They are notably not the same as build deps. Just want to make sure we don't add bloat, both to the compiled binaries and deps that get pulled in at build time.

riedgar-ms: According to:
https://doc.rust-lang.org/rust-by-example/testing/dev_dependencies.html
they're where you put things for testing and are not propagated.

hudson-ai:
Just checking my mental model: 
- we need an actual parser object to do these tests, and previously we had one of those in sample parser
- we now also have one of those in the test utils (do those now essentially operate as an additional sample parser PLUS testing utility functions? i.e. would the DRY thing be to take sample_parser as a dependency in test_utils? not saying to do that...)
- sample_parser is preserved as a reference usage for users rather than as a place-where-tests-live
- test_utils takes llguidance (i.e. parser) as a dependency, but llguidance/parser takes test utils as a dev dependency? i.e. it is in some way just #[cfg(test)] in parser? (being split in its own crate actually feels good though)

riedgar-ms:
> Just checking my mental model:
> 
> * we need an actual parser object to do these tests, and previously we had one of those in sample parser
> * we now also have one of those in the test utils (do those now essentially operate as an additional sample parser PLUS testing utility functions? i.e. would the DRY thing be to take sample_parser as a dependency in test_utils? not saying to do that...)
> * sample_parser is preserved as a reference usage for users rather than as a place-where-tests-live
> * test_utils takes llguidance (i.e. parser) as a dependency, but llguidance/parser takes test utils as a dev dependency? i.e. it is in some way just #[cfg(test)] in parser? (being split in its own crate actually feels good though)

1) Yes
2) Sort of.... there is the contents of `lib.rs` which are a quick way of getting an actual parser (basically loading in the TokEnv). There are only a couple of lines (sort of) duplicated, though. The parser is being created around lines 147 of `sample_parser.rs`
3) That was the major goal
4) Yes.... it is a bit weird, but I'm not sure there's a better way (given that json_schema_test_suite also wants some of the utilities). Also, being very technical, I think that it's the `llguidance/parser/test` crate which takes the `llg_test_utils` crate as a dependency, not `llguidance/parser` (which would be a circular dependency)

I do plan a follow up to move `XorShift` out of `parser` and into `llg_test_utils`. Although rather than have `sample_parser` use it as a dependency, I'll have it use the regular `rand` crate (since my 'performance' speculation cannot apply to that)

hudson-ai (approved):
I think that the reorganization here is worthwhile, even if we still have a complex dependency graph. At the moment, this does feel unavoidable given that llguidance was written against a trait rather than a concrete type. LGTM

riedgar-ms:
Just pushed another update after noticing that the JSON checking routines in `json_schema_test_suite` are almost identical to those in `llg_test_utils`. This does simplify the dependencies somewhat, but there is still a 'circular' dependency in the sense that `llg_test_utils` has a direct dependency on `llguidance`, while `llguidance` has a dev-dependency on `llg_test_utils`

PR #321, opened 2026-03-27
fix: correctly intersect patternProperties with properties and additionalProperties

hudson-ai:
## Summary

Fixes two bugs where `patternProperties` constraints were silently dropped:

1. **Within a schema**: A named property matching a pattern property regex was not intersected with the pattern's schema. Per the JSON Schema spec (§10.3.2), a property must validate against *both* its `properties` schema and any matching `patternProperties` schema.

2. **Across schemas** (e.g., `allOf`): When intersecting two object schemas, pattern properties that existed only on one side were carried as-is without being intersected with the other side's `additionalProperties`. From that other schema's perspective, properties matching the pattern are "additional" and must satisfy its `additionalProperties` constraint.

## Changes

- **`schema.rs`** (`compile_object`): Pre-intersect each named property with any matching pattern property schema before assembly.
- **`schema.rs`** (`intersect`): When merging pattern properties during object intersection, intersect one-sided patterns with the other side's `additionalProperties` (defaulting to `Schema::Any` if absent).
- **`shared_context.rs`**: Add `property_schema_matches` helper for regex matching.
- **`test_lark.rs`**: Comprehensive tests covering all pattern property interactions — satisfiable intersections, unsatisfiable intersections, required properties, cross-schema with `additionalProperties: false`, and same-key/different-key pattern merging via `allOf`.

## Maskbench

`JsonSchemaStore---theme.json` changes error from "oneOf constraints are not supported" to "Schema intersection stack level exceeded". The schema has 98 named properties all matching a single pattern property, each `$ref`ing the same complex definition. The pre-intersection now triggers before the oneOf check, but the schema fails to compile either way. Baseline updated.

## Test suite ratchet

1 improvement:
```
patternProperty invalidates property: false_positive → pass
```

### DIFF
--- a/json_stats/expected_maskbench.json
+++ b/json_stats/expected_maskbench.json
@@ -11802,7 +11802,7 @@
     "json_error": "oneOf constraints are not supported. Enable 'coerce_one_of' option to approximate oneOf with anyOf\n  while processing https://json.schemastore.org/testenvironments.json#/definitions/config"
   },
   "JsonSchemaStore---theme.json": {
-    "json_error": "oneOf constraints are not supported. Enable 'coerce_one_of' option to approximate oneOf with anyOf"
+    "json_error": "Schema intersection stack level exceeded"
   },
   "JsonSchemaStore---tikibase.schema.json": {},
   "JsonSchemaStore---tizen_workspace.json": {},
--- a/sample_parser/tests/test_lark.rs
+++ b/sample_parser/tests/test_lark.rs
@@ -853,6 +853,32 @@ fn test_json_pattern_properties() {
         ],
     );
 
+    // Non-trivial intersection: "count" is declared as integer via properties,
+    // and matches "^c" pattern which requires multipleOf: 5. The intersection
+    // should produce integer ∩ multipleOf(5) = integers that are multiples of 5.
+    json_test_many(
+        &json!({
+            "type": "object",
+            "properties": {
+                "count": { "type": "integer", "minimum": 0 },
+            },
+            "patternProperties": {
+                "^c": { "type": "integer", "multipleOf": 5 },
+            },
+            "required": ["count"],
+        }),
+        &[
+            json!({"count": 0}),
+            json!({"count": 10}),
+            json!({"count": 25}),
+        ],
+        &[
+            json!({"count": 3}),
+            json!({"count": 7}),
+            json!({"count": -5}),
+        ],
+    );
+
     // Same schema but "foo" is required — since "foo" is unsatisfiable
     // (string ∩ integer), the entire schema is unsatisfiable.
     json_err_test(
--- a/sample_parser/tests/test_lark.rs
+++ b/sample_parser/tests/test_lark.rs
@@ -872,6 +872,40 @@ fn test_json_pattern_properties() {
         }),
         "required property 'foo' is unsatisfiable",
     );
+
+    // Cross-schema: patternProperties from one side must be intersected with
+    // additionalProperties from the other side. Here, the first schema says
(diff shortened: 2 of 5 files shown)


sample_parser/tests/test_lark.rs
```diff
@@ -800,6 +800,9 @@ fn test_json_pattern_properties() {
         "required property 'foo' is unsatisfiable",
     );
 
+    // "foo" matches patternProperties "^foo" (type: integer), but properties says
```
riedgar-ms, line 803: Just wondering... don't these tests really belong under the JSON tests? Or is there something about them that needs the lark 'wrapper' around them?

json_stats/expected_maskbench.json
```diff
@@ -11802,7 +11802,7 @@
     "json_error": "oneOf constraints are not supported. Enable 'coerce_one_of' option to approximate oneOf with anyOf\n  while processing https://json.schemastore.org/testenvironments.json#/definitions/config"
   },
   "JsonSchemaStore---theme.json": {
-    "json_error": "oneOf constraints are not supported. Enable 'coerce_one_of' option to approximate oneOf with anyOf"
+    "json_error": "Schema intersection stack level exceeded"
```
riedgar-ms, line 11805: Is this something the user can fix by changing a constant and recompiling?

sample_parser/tests/test_lark.rs
hudson-ai: Yes, putting these with the other json tests would be more sensible. For whatever reason, these existing json tests haven't been moved yet though.

json_stats/expected_maskbench.json
hudson-ai: `parser/src/json/schema.rs::SchemaBuilderOptions::max_stack_level` has a default of 128 and isn't currently user-configurable. The issue is that setting this number too high can open users up to *actual* stack overflows due to some inefficient recursion that I wrote here.

I actually have a draft rewrite on my local machine that unrolls the recursion into something explicitly iterative, so I'd actually like to avoid exposing this control (as I want to get rid of the very thing that makes it necessary...). I'll try to get that in as a draft PR for you to look at before I go on holiday :)

riedgar-ms: Some semi-formed thoughts:

1. Will you put an iteration cap in as well? Unless you can make this completely iterative, managing your own stack is just going to mean something overly complex can run away for longer before getting OoM-killed (not to say that going iterative is the wrong thing to do)
2. Related to the above, can we give the users actionable error messages? Saying "`oneOf` not supported" is actually useful (at least if the user can rewrite the schema). "Schema intersection stack level exceeded" is not. I suppose I'm asking a more meta-question of whether we can point users to other ways of writing their schema which might be less resource intensive

hudson-ai: Good questions
1. Yes, but in my experience, recursion depth limits are much much easier to hit than OOM limits. I'd be okay exposing a control on that -- Michal already uses the notion of "fuel" elsewhere
2. Really hard to say. We hit the limit, but it's not trivial to trace the root cause

This being said, I would say that all of this is out of scope for *this* PR. There are existing "Schema intersection stack level exceeded" failures, and it's more or less incidental that this particular change flipped this case over to that camp. Happy to dig deeper if you disagree though.

riedgar-ms: Absolutely, it takes longer to exhaust the heap than the stack :-) That's a potential concern in another way, but I'm happy for it to be out of scope for this PR.

As for the error message, it's something we should probably think about, but it can be out of scope for this PR. If this lets us succeed on some schema where we failed before, I've no objections to merging.

PR #324, opened 2026-03-31
[Test] Suppress `json_schema_test_suite` output

riedgar-ms:
Since the refactoring to a common function for checking JSON schema behaviour, the `json_schema_test_suite` has become markedly more verbose. Add an extra flag in the `llg_test_utils` to allow this to be suppressed.


llg_test_utils/src/lib.rs
```diff
@@ -23,13 +23,30 @@
...
 use lazy_static::lazy_static;
 use llguidance::{
     earley::SlicedBiasComputer,
     toktrie::{InferenceCapabilities, TokEnv},
     ParserFactory,
 };
 
+// ── Global quiet-mode flag ───────────────────────────────────────────────────
+
+static QUIET_MODE: AtomicBool = AtomicBool::new(false);
```
riedgar-ms, line 37: I'm not overjoyed at this global, but it appears that Rust doesn't have default arguments. As such the alternatives are making `json_schema_check()` into a wrapper, changing all of the calls to `json_schema_check()` to add a `verbose` parameter, or doing this.

riedgar-ms: If anyone feels strongly about it, I'm happy to ~~make the changes~~ have CoPilot change the code.

nking-1: Would an environment variable suffice instead?

riedgar-ms: I suppose it would, but within the test utilities themselves, I'd prefer the global over that.

nking-1 (approved):

PR #334, opened 2026-04-14
Add llg_get_version() binding

nking-1:
Adds a binding to dump the llguidance version, along with key dependency versions. Adding this binding also makes sure the version itself is stored as a string in the built binary files, so we can dump it using `strings` later on.

This change depends on the following derivre PR: https://github.com/guidance-ai/derivre/pull/11

### DIFF
--- a/python/llguidance/__init__.py
+++ b/python/llguidance/__init__.py
@@ -14,6 +14,10 @@ from ._tokenizer import TokenizerWrapper
 from ._grammar_from import GrammarFormat, grammar_from
 from ._struct_tag import StructTag
 
+from importlib.metadata import version as _pkg_version
+
+__version__ = _pkg_version("llguidance")
+
 __all__ = [
     "LLTokenizer",
     "LLMatcher",
@@ -29,4 +33,5 @@ __all__ = [
     "StructTag",
     "regex_to_lark",
     "get_version",
+    "__version__",
 ]
--- a/python/llguidance/__init__.py
+++ b/python/llguidance/__init__.py
@@ -8,6 +8,7 @@ from ._lib import (
     LLMatcher,
     LLParserLimits,
     regex_to_lark,
+    get_version,
 )
 from ._tokenizer import TokenizerWrapper
 from ._grammar_from import GrammarFormat, grammar_from
@@ -27,4 +28,5 @@ __all__ = [
     "GrammarFormat",
     "StructTag",
     "regex_to_lark",
+    "get_version",
 ]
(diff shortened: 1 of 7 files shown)


riedgar-ms:
Looks like this needs the `derivre` change merged and released?

Also, should there be changes in `ffi.rs` and the Python bindings?

nking-1:
> Looks like this needs the `derivre` change merged and released?
> 
> Also, should there be changes in `ffi.rs` and the Python bindings?

Yes, this depends on the `derivre` change first. The `ffi.rs` changes alone were enough to make it possible to dump with `strings`, so I didn't add Python bindings yet. I could add that though if we want a way to dump version info from Python.

nking-1:
Added Python bindings and tested:

guidance % .venv/bin/python -c "import llguidance; print(llguidance.get_version())"
llguidance@1.7.2 derivre@0.3.11

python/llguidance/_lib.pyi
```diff
@@ -715,3 +715,6 @@ def regex_to_lark(regex: str, use_ascii: str = "d") -> str:
     For Python2 or byte patters in Python3 semantics `use_ascii = "dws"`
     More flags may be added in future.
     """
+
+def get_version() -> str:
```
hudson-ai, line 719: Is there a way to expose this as a static `__version__` string on the package?

nking-1: Change made, thank you for the suggestion. Here's the way it will work, with output:

python -c "import llguidance; print(llguidance.__version__)"
llguidance@1.7.2 derivre@0.3.11

python/llguidance/__init__.py
```diff
@@ -8,11 +8,14 @@
     LLMatcher,
     LLParserLimits,
     regex_to_lark,
+    get_version,
 )
 from ._tokenizer import TokenizerWrapper
 from ._grammar_from import GrammarFormat, grammar_from
 from ._struct_tag import StructTag
 
+__version__ = get_version()
```
hudson-ai, line 17: Python's `__version__` should be [PEP 440](https://peps.python.org/pep-0440/) compliant, i.e. just `1.7.2` or whatever

nking-1: Good catch, changed to: `__version__ = _pkg_version("llguidance")` which will dump from the metadata directly

So now we have:
`print(repr(llguidance.__version__))` outputs 1.7.2
`print(repr(llguidance.get_version()))` outputs llguidance@1.7.2 derivre@0.3.11

Hopefully that won't be confusing.

hudson-ai: Looks great!

hudson-ai (approved):
Looks good to me :)

PR #341, opened 2026-04-23
More clippy linter fixes + Rust toolchain version pin

nking-1:
Fixes a few more linter errors that I missed in the last round. This one should cover the rest of them.

Also pins rust toolchain version using rust-toolchain.toml as documented here:

https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

Components selected from what the CI uses:

 - clippy: lint.yml runs cargo clippy
 - rustfmt: lint.yml runs cargo fmt --check


riedgar-ms (approved):
No objections, although I may not be best placed...

PR #342, opened 2026-04-23
Pin Rust toolchain to 1.95.0 via rust-toolchain.toml

nking-1:
Pins rust toolchain version using rust-toolchain.toml as documented here:

https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

Components selected from what the CI uses:

 - clippy: lint.yml runs cargo clippy
 - rustfmt: lint.yml runs cargo fmt --check

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,3 @@
+[toolchain]
+channel = "1.95.0"
+components = ["clippy", "rustfmt"]


nking-1:
Closing - looks like we have to do the linter fix + version pin in 1 PR to make the CI handle it cleanly

hudson-ai:
Is the `rustc` version used when building and running tests unpinned, or does this pin that too? Side note, but we also have a pin on 1.87.0 for the MSRV check.

PR #345, opened 2026-04-28
[Build] New workflow for releases

riedgar-ms:
Create a new workflow for performing releases. This should:

- Maximally reuse existing workflows
- Avoid split releases due to crates.io updating completely independently of pypi

Leave the existing `wheels.yml` in place, but turn make it workflow_dispatch only. Once we're confident the new workflow is operating correctly, `wheels.yml` can be removed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,9 @@
 name: Release
 
 on:
+  push:
+    tags:
+      - 'v*'
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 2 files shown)


hudson-ai:
Why leave the existing wheels.yml?

riedgar-ms:
> Why leave the existing wheels.yml?

For reference until this new workflow has actually done a release. It does need to go, but it's easier to compare if you're not flipping back and forth in version control. I've updated the PR description

hudson-ai (approved):

PR #346, opened 2026-04-29
Simplify release workflow graph

hudson-ai:
## Summary
- Remove the old standalone wheels workflow.
- Add explicit validate and build fan-in jobs to the release workflow.
- Simplify publish job dependencies so both publish jobs depend on the build fan-in.

## Validation
- `git diff --check`
- Parsed `.github/workflows/release.yml` as YAML


.github/workflows/release.yml
```diff
@@ -195,14 +202,21 @@ jobs:
           name: release-macos-${{ matrix.platform.target }}-${{ matrix.python-version }}
           path: dist
 
+  build:
+    name: Build
```
riedgar-ms: Perhaps 'End of Build' ?

.github/workflows/release.yml
```diff
@@ -43,13 +43,20 @@ jobs:
       python-version: '3.12'
       ref: ${{ github.sha }}
 
+  validate:
+    name: Validate
```
riedgar-ms: Minor thing but 'End of Validation' ?

riedgar-ms (approved):
Some minor naming thoughts

PR #347, opened 2026-05-06
Fix validate_tokens for multiple EOS tokens

fellhorn:
`validate_tokens` in `parser/src/earley/parser.rs` only recognized the primary EOS token, which was overlooked in #305 when adding support for multiple EOS tokens.

This PR aligns the behaviour with e.g. `compute_mask` that now allows multiple eos tokens.


hudson-ai:
Awesome, thanks @fellhorn for correcting my oversight!

PR #352, opened 2026-06-11
Improve FFI safety

riedgar-ms:
## FFI Safety Improvements

Harden the C FFI layer (`parser/src/ffi.rs`, `parser/src/ffi_par.rs`) against undefined behavior, with  comprehensive test coverage and updated documentation.

### Key Changes

**Panic safety at FFI boundaries**
- Add `ffi_guard_ptr` and `ffi_guard_constraint` helpers that wrap all `extern "C"` entry points with  `catch_unwind`, preventing Rust panics from unwinding into C callers.
- Use `panic_utils::catch_unwind` (suppresses stderr noise, captures backtraces) instead of  `std::panic::catch_unwind`.
- Replace `assert!()` / `unwrap()` in rayon workers with per-constraint error reporting + outer `catch_unwind`  safety net.

**Null-pointer hardening**
- All `llg_free_*` functions now accept null (matching the `free(NULL)` idiom).
- Null `step.constraint` in parallel compute is handled gracefully (skip) instead of dereferencing.
- `llg_par_compute_mask` with null `steps` and `n_steps > 0` no longer invokes the done callback.

**API documentation & header improvements**
- `# Safety` docs added to all `unsafe impl` blocks, helper functions, and `extern "C"` functions.
- `LLGUIDANCE_NODISCARD` macro added for `llg_compute_mask` and `llg_commit_token` (C++17 `[[nodiscard]]` / GCC  `warn_unused_result` / MSVC `_Check_return_`).
- Document zero-init requirement for `LlgConstraintInit`.
- Document async semantics of `llg_par_compute_mask` (callback threading, lifetime requirements).
- Remove stale `NodeRef` from generated header.

**Internal cleanup**
- `make_c_string()` helper centralizes NUL-terminated string creation (replaces fragile manual `\0` appending).
- `SendPtr` newtype in `ffi_par.rs` replaces ad-hoc `CbData` struct.
- Read-only matcher functions take `&` instead of `&mut`.

**Test coverage (C++)**
- Null-free safety for all 4 handle types (tokenizer, constraint, matcher, stop_controller).
- Matcher error propagation from failed `consume`/`rollback`.
- Parallel compute: null callback, null steps, null constraint (solo + mixed batch), error propagation.
- NUL-termination of error strings, log output, and stop controller commit output.
- New `test_parallel.cpp` file with dedicated parallel-compute tests.

**Other**
- Regenerated `llguidance.h` via cbindgen.
- Added `CONTRIBUTING.md` with development commands and header regeneration instructions.


riedgar-ms:
Looks like I need to rerun cbindgen, but would appreciate thoughts.

riedgar-ms:
This should wait on #353 anyway

parser/src/ffi.rs
```diff
@@ -829,29 +923,62 @@ pub extern "C" fn llg_commit_token(
...
+/// immediately and the function returns. If the `rayon` feature is not enabled,
+/// the callback is invoked immediately and no work is done.
+///
 /// # Safety
-/// This function should only be called from C code.
+/// - `steps` must point to an array of `n_steps` valid [`LlgConstraintStep`]
+///   elements (or be null, in which case this is a no-op).
+/// - Each step's `constraint` pointer must remain valid and unaliased for the
+///   duration of the parallel computation.
+/// - Each step's `mask_dest` must point to a buffer of at least
```
mmoskal, line 972: **[P1] Require destination buffers to be disjoint**

This contract only requires each `mask_dest` to be large enough. Rayon writes different steps concurrently, so two steps can satisfy the documented contract while using the same or overlapping destination ranges, which creates a data race and undefined behavior. Require all `mask_dest[..mask_byte_len]` regions to be pairwise non-overlapping until completion.

parser/src/ffi.rs
```diff
@@ -829,29 +923,62 @@ pub extern "C" fn llg_commit_token(
...
+///   `mask_byte_len` bytes.
+/// - `user_data` is passed opaquely to `done_cb`; the caller must ensure it
```
mmoskal, line 977: **[P1] Document callback thread-safety requirements**

With `done_cb` present, `ffi_par` invokes it on a Rayon worker while the original caller may continue. Keeping `user_data` alive is not sufficient: the callback and any state it accesses must be safe for cross-thread, potentially concurrent access. Document the worker-thread invocation and synchronization requirement; `extern "C" fn` alone does not provide thread safety.

parser/src/ffi.rs
```diff
@@ -829,29 +923,62 @@ pub extern "C" fn llg_commit_token(
...
 pub unsafe extern "C" fn llg_par_compute_mask(
     steps: *const LlgConstraintStep,
     n_steps: usize,
     user_data: *const c_void,
     done_cb: LlgCallback,
 ) {
-    if steps.is_null() {
-        panic!("llg_par_compute_mask: steps is null");
+    // Gracefully handle null/empty — invoke callback and return.
+    if steps.is_null() || n_steps == 0 {
```
mmoskal: **[P2] Do not report a null nonempty batch as completed**

`n_steps == 0` is a valid no-op, but `steps == NULL` with `n_steps > 0` violates the pointer/length invariant. Invoking `done_cb` here makes that invalid call indistinguishable from successful mask computation, and there is no constraint available on which to record an error, so callers can consume stale masks. Keep the empty-batch case, but fail loudly for a null nonempty batch.

parser/src/ffi.rs
```diff
@@ -37,17 +38,75 @@ use toktrie::{
...
+/// returns null.
+///
+/// Use for functions like `llg_new_tokenizer` that report errors via a
+/// caller-provided buffer.
+fn ffi_guard_ptr<T>(
+    f: impl FnOnce() -> *mut T + std::panic::UnwindSafe,
+    error_string: *mut c_char,
+    error_string_len: usize,
+) -> *mut T {
+    match std::panic::catch_unwind(f) {
```
mmoskal: **[P2] Use the crate panic guard**

`std::panic::catch_unwind` runs the process panic hook before returning `Err`, so panics converted into FFI errors here still print to the host application's stderr. The existing `panic_utils::catch_unwind` installs the intended hook behavior and captures the backtrace for the returned error. Use it consistently for these new guards and the direct matcher/validation guards.

mmoskal:
I like the direction of changes, some comments from the robot above!

riedgar-ms:
I think that's all the comments addressed @mmoskal ?

parser/src/ffi.rs
mmoskal: the principal FFI guards now use panic_utils::catch_unwind, but the parallel worker still uses std::panic::catch_unwind. Worker panics can therefore still invoke the host panic hook and print to stderr. parser/src/ffi_par.rs (lines 34-36)

parser/src/ffi.rs
mmoskal: maybe better to just have it panic not return?

parser/src/ffi.rs
```diff
@@ -837,29 +951,78 @@ pub extern "C" fn llg_commit_token(
...
 pub unsafe extern "C" fn llg_par_compute_mask(
     steps: *const LlgConstraintStep,
     n_steps: usize,
     user_data: *const c_void,
     done_cb: LlgCallback,
 ) {
+    // Empty batch is a valid no-op — invoke callback and return.
+    if n_steps == 0 {
+        if let Some(cb) = done_cb {
+            cb(user_data);
```
mmoskal: ideally this would be invoked on a rayon thread, or at least the docs updated to say it will be invoked on the same thread when n_steps=0; this can lead to deadlocks in some cases

mmoskal (approved):

riedgar-ms:
@mmoskal Just had another pass at your comments, and I think they're all addressed.

There are even more suggestions pending from Copilot, but this is starting to get large and unwieldly. Unless you feel there's something missing from this PR, I'd like to get it in, and look at the next set of suggestions in a new PR.

parser/src/ffi.rs
```diff
@@ -981,49 +985,54 @@ pub unsafe extern "C" fn llg_par_compute_mask(
...
-    if steps.is_null() {
-        return;
-    }
+    // Wrap the entire body in catch_unwind so that panics (e.g. from the
+    // assert below) don't abort the host process across the FFI boundary.
+    let _ = panic_utils::catch_unwind(AssertUnwindSafe(|| {
+        assert!(
+            !steps.is_null() || n_steps == 0,
+            "llg_par_compute_mask: steps is null but n_steps > 0"
+        );
```
mmoskal: the assert right now will just be ignored; it should be outside the catch_unwind

mmoskal (approved):
please move the assert outside, otherwise good to merge!

parser/src/ffi.rs
riedgar-ms: Since I'm not really familiar with this code, I'd just like to clarify what needs to happen here. According to my bot:

> The pattern follows best practice for preventing UB (panic never crosses FFI) but violates the function's own completion contract. The let _ = discard, combined with `done_cb` being owned by the spawned task, means any panic before the spawn is both silently ignored and leaves async callers deadlocked

For this, I do see that the 'non-rayon' side will "invoke callback to avoid deadlocking the caller.'

The bot is proposing a pattern like:

```rust
let prep = panic_utils::catch_unwind(AssertUnwindSafe(|| {
    if steps.is_null() && n_steps != 0 {
        bail!("steps is null but n_steps > 0");
    }
    Ok(if n_steps == 0 {
        Vec::new()
    } else {
        unsafe { std::slice::from_raw_parts(steps, n_steps).to_vec() }
    })
}));

match prep {
    Ok(steps) => {
        // par_compute_mask now owns done_cb and guarantees it fires.
        crate::ffi_par::par_compute_mask(steps, user_data, done_cb);
    }
    Err(_e) => {
        // Point-of-no-return not reached: we still own done_cb, fire it once.
        if let Some(cb) = done_cb {
            cb(user_data);
        }
    }
}
```

Does that look correct?

riedgar-ms: Side note, it is also proposing to harden `par_compute_mask()` by ensuring that the callback is always invoked.

riedgar-ms: @mmoskal I've had it go ahead and make those changes. Please let me know if that wasn't what you meant.

mmoskal (approved):
LGTM!

PR #353, opened 2026-06-11
Add C++ FFI tests

riedgar-ms:
Comprehensive test suite for `parser/src/ffi.rs` covering:
- Tokenizer creation (v1, v2, clone, error cases)
- Tokenization functions (bytes, marker, stringify, decode)
- Constraint lifecycle (all constraint types, full sampling loops)
- Matcher API (mask, consume, rollback, reset, validate, ff_tokens)
- Stop controller (tokens, regex, clone)
- Parallel compute_mask (rayon)
- Version string

A buffer-sizing bug in `ffi.rs` was also detected and fixed (and a corresponding test added).

Build system: CMake with Boost.Test, linking libllguidance statically.
CI: New c-ffi-tests.yml workflow, integrated into build-all.yml.

### DIFF
--- a/c_ffi_tests/src/test_matcher.cpp
+++ b/c_ffi_tests/src/test_matcher.cpp
@@ -152,30 +152,40 @@ BOOST_AUTO_TEST_CASE(is_stopped_when_forced_eos) {
 
 BOOST_AUTO_TEST_CASE(rollback) {
   MatcherContext ctx;
-  auto matcher = ctx.make_matcher("regex", "[a-z]+");
+  // "ab|ac" distinguishes states: after 'a', both 'b' and 'c' are valid.
+  // Consuming 'a','b' commits to the "ab" branch; rolling back 1 restores
+  // the post-'a' state where 'c' is also reachable.
+  auto matcher = ctx.make_matcher("regex", "ab|ac");
 
   check_matcher_has_no_error(matcher.get());
-  BOOST_REQUIRE_EQUAL(llg_matcher_consume_token(matcher.get(), 97), 0);
-  BOOST_REQUIRE_EQUAL(llg_matcher_consume_token(matcher.get(), 98), 0);
+  BOOST_REQUIRE_EQUAL(llg_matcher_consume_token(matcher.get(), 97), 0); // 'a'
+  BOOST_REQUIRE_EQUAL(llg_matcher_consume_token(matcher.get(), 98), 0); // 'b'
   BOOST_REQUIRE_EQUAL(llg_matcher_rollback(matcher.get(), 1), 0);
-  BOOST_REQUIRE_EQUAL(llg_matcher_consume_token(matcher.get(), 99), 0);
-  BOOST_CHECK(!llg_matcher_is_error(matcher.get()));
+
+  // After rollback we should be back in the post-'a' state
+  BOOST_REQUIRE_EQUAL(llg_matcher_compute_mask(matcher.get()), 0);
+  const uint32_t *mask = llg_matcher_get_mask(matcher.get());
+  BOOST_REQUIRE(mask != nullptr);
+  BOOST_CHECK(mask_has_token(mask, 99));  // 'c' should be allowed
+  BOOST_CHECK(mask_has_token(mask, 98));  // 'b' should be allowed
+  BOOST_CHECK(!mask_has_token(mask, 48)); // '0' should be rejected
 }
 
 BOOST_AUTO_TEST_CASE(reset) {
   MatcherContext ctx;
-  auto matcher = ctx.make_matcher("regex", "[a-z]+");
-  const uint32_t tokens[] = {97, 98, 99};
+  // "ab" requires exactly 'a' then 'b'. After consuming 'a', only 'b' is valid.
+  // After reset, only 'a' (the start) should be valid, not 'b'.
+  auto matcher = ctx.make_matcher("regex", "ab");
 
   check_matcher_has_no_error(matcher.get());
-  BOOST_REQUIRE_EQUAL(
-      llg_matcher_consume_tokens(matcher.get(), tokens, std::size(tokens)), 0);
+  BOOST_REQUIRE_EQUAL(llg_matcher_consume_token(matcher.get(), 97), 0); // 'a'
   BOOST_REQUIRE_EQUAL(llg_matcher_reset(matcher.get()), 0);
   BOOST_REQUIRE_EQUAL(llg_matcher_compute_mask(matcher.get()), 0);
 
   const uint32_t *mask = llg_matcher_get_mask(matcher.get());
   BOOST_REQUIRE(mask != nullptr);
-  BOOST_CHECK(mask_has_token(mask, 97));
+  BOOST_CHECK(mask_has_token(mask, 97));  // 'a' allowed (start position)
+  BOOST_CHECK(!mask_has_token(mask, 98)); // 'b' not allowed (would require 'a' first)
 }
 
 BOOST_AUTO_TEST_CASE(validate_tokens) {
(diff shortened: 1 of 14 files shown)


c_ffi_tests/src/test_matcher.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+  check_matcher_has_no_error(matcher.get());
+  BOOST_CHECK_EQUAL(
+      llg_matcher_validate_tokens(matcher.get(), tokens, std::size(tokens)), 1);
+}
+
+BOOST_AUTO_TEST_CASE(compute_ff_tokens) {
+  MatcherContext ctx;
+  auto matcher = ctx.make_matcher("regex", "hello");
+  std::vector<uint32_t> output(5);
```
mmoskal, line 214: **[P1] Cover short fast-forward buffers**

This exact-sized buffer misses an existing out-of-bounds write: `llg_matcher_compute_ff_tokens` calculates `len = min(v.len(), output_len)` but copies `v.len()` elements. A smaller C buffer is overrun. Add a short-buffer/canary case and fix the Rust copy count to `len`.

c_ffi_tests/src/test_tokenizer.cpp
```diff
@@ -0,0 +1,139 @@
...
+  char error_buf[256] = {};
+  LlgTokenizer *tok = llg_new_tokenizer_v2(nullptr, error_buf, sizeof(error_buf));
+
+  BOOST_TEST(tok == nullptr);
+  BOOST_TEST(std::strcmp(error_buf, "tok_init is NULL") == 0);
+}
+
+BOOST_AUTO_TEST_CASE(tokenizer_v2_eos_out_of_range) {
+  SmallVocab vocab;
+  uint32_t extra_eos[] = {99};
```
mmoskal, line 99: **[P1] Reject an invalid primary EOS**

This checks only an invalid extra EOS. With no extra list, both tokenizer APIs currently accept an out-of-range primary `tok_eos`; a later mask computation can index that token and panic across the C boundary. Add primary-EOS cases for v1 and v2, and validate the primary ID unconditionally.

c_ffi_tests/src/test_parallel.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+void wait_for_done(const std::atomic<bool> &done) {
+  const auto deadline =
+      std::chrono::steady_clock::now() + std::chrono::seconds(2);
+  while (!done.load(std::memory_order_acquire) &&
+         std::chrono::steady_clock::now() < deadline) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
+  }
+  BOOST_REQUIRE(done.load(std::memory_order_acquire));
```
mmoskal: **[P2] Timeout creates use-after-free**

If the two-second deadline expires, `BOOST_REQUIRE` unwinds while the Rayon task can still access the stack-owned atomic, mask buffers, and constraints through copied pointer descriptors. The test must not return until the callback has completed; otherwise its failure path has undefined behavior.

c_ffi_tests/src/test_matcher.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+BOOST_AUTO_TEST_CASE(reset) {
+  MatcherContext ctx;
+  auto matcher = ctx.make_matcher("regex", "[a-z]+");
+  const uint32_t tokens[] = {97, 98, 99};
+
+  check_matcher_has_no_error(matcher.get());
+  BOOST_REQUIRE_EQUAL(
+      llg_matcher_consume_tokens(matcher.get(), tokens, std::size(tokens)), 0);
+  BOOST_REQUIRE_EQUAL(llg_matcher_reset(matcher.get()), 0);
```
mmoskal, line 182: **[P2] Rollback and reset can be no-ops**

Both tests still pass if the tested operation does nothing: `[a-z]+` accepts the subsequent inputs from either state. Use state-distinguishing grammars, such as `ab|ac` for rollback and `ab` for reset, and assert both an allowed and rejected next token.

c_ffi_tests/src/test_tokenize.cpp
```diff
@@ -0,0 +1,150 @@
...
+      tok.tok, reinterpret_cast<const uint8_t *>(input.data()), input.size(),
+      tokens.data(), tokens.size());
+
+  BOOST_REQUIRE_EQUAL(n, 5u);
+  check_tokens(tokens, {104, 101, 108});
+}
+
+BOOST_AUTO_TEST_CASE(tokenize_bytes_marker) {
+  ByteTokenizer tok;
+  const std::string input = "hello";
```
mmoskal: **[P2] Marker path is not exercised**

The input contains no `0xff` marker, so this follows ordinary byte tokenization and cannot catch regressions in marker parsing. Include a numeric marker such as the byte sequence for `0xff[256]` and verify it produces the EOS token.

c_ffi_tests/src/test_stop_controller.cpp
```diff
@@ -0,0 +1,165 @@
...
+  size_t output_len = 0;
+  bool is_stopped = true;
+  BOOST_TEST(committed_text(stop_ctrl.get(), 104, output_len, is_stopped) ==
+             "h");
+  BOOST_TEST(is_stopped == false);
+  BOOST_TEST(committed_text(stop_ctrl.get(), 101, output_len, is_stopped) ==
+             "e");
+  BOOST_TEST(is_stopped == false);
+
+  StopControllerPtr clone(llg_clone_stop_controller(stop_ctrl.get()));
```
mmoskal, line 149: **[P2] Clone test has no state to clone**

This controller uses only an EOS stop token. After committing `he`, its relevant state is equivalent to a fresh controller, so an implementation that creates a new controller instead of cloning would pass. Clone midway through a regex stop sequence, then complete it independently on both controllers.

parser/src/ffi.rs
```diff
@@ -171,7 +171,14 @@ impl LlgTokenizer {
             token_bytes
         };
 
-        let mut trie = TokTrie::from(&TokRxInfo::new(tokens.len() as u32, init.tok_eos), &tokens);
+        let vocab_size = tokens.len() as u32;
+        ensure!(
```
mmoskal, line 176: **[P1] Preserve the no-EOS sentinel**

`TokTrie` and the parser intentionally support `INVALID_TOKEN` (`UINT32_MAX`) as the no-EOS sentinel; downstream EOS paths explicitly check for it before indexing the vocabulary. This unconditional range check rejects that previously supported configuration. Allow `init.tok_eos == INVALID_TOKEN` while continuing to reject other IDs outside the vocabulary.

c_ffi_tests/src/test_tokenize.cpp
```diff
@@ -0,0 +1,247 @@
...
+  BOOST_CHECK_EQUAL(n, 1u);
+}
+
+// NOTE: bare \xff handling (not followed by '[' or '<') is currently undefined
+// behavior in llg_tokenize_bytes_marker — the byte may be silently dropped.
+// A future fix should tokenize it literally as token 255.
+
+BOOST_AUTO_TEST_CASE(decode_tokens_include_special) {
+  ByteTokenizer tok;
+  const uint32_t token = BYTE_TOK_EOS;
```
mmoskal, line 194: **[P2] Exercise an actually special token**

`create_byte_tokenizer()` stores `BYTE_TOK_EOS` as ordinary bytes `<EOS>` without the `0xff` special-token prefix. `decode_ext` therefore emits it regardless of `LLG_DECODE_INCLUDE_SPECIAL`, so these tests pass even if the flag is ignored. Add a genuinely marker-prefixed special token and compare decoding with and without the flag.

c_ffi_tests/src/test_tokenizer.cpp
```diff
@@ -0,0 +1,234 @@
...
+  BOOST_REQUIRE_MESSAGE((tok != nullptr),
+                        "Expected tokenizer creation to succeed: " << error_buf);
+  llg_free_tokenizer(tok);
+}
+
+BOOST_AUTO_TEST_CASE(tokenizer_v1_vocab_size_zero) {
+  LlgTokenizerInit tok_init = {};
+  tok_init.vocab_size = 0;
+  tok_init.tok_eos = 0;
+  tok_init.token_lens = nullptr;
```
mmoskal: **[P2] Reach the EOS validation this test names**

With both vocabulary pointers null, tokenizer construction fails at `token_lens and token_bytes must be set` before reaching the new zero-vocabulary/EOS range check. The test therefore passes for the wrong reason. Supply non-null dummy pointers, which are not dereferenced for a zero-length vocabulary, and assert that the returned error is the EOS out-of-range error.

mmoskal (commented):
thank you for addressing previous comments! some more below... the EOS = INVALID_TOKEN seems like a regression (not sure why we added this, but I'm sure there was a reason)

riedgar-ms:
@mmoskal I think that addresses the three points raised?

mmoskal (approved):
Thank you!

hudson-ai (approved):

Issue #54, opened 2024-11-14
lark grammar doesn't support \x00 and similar in string literals


mmoskal:
I think the python lark supports it (though would be good to check!)

To fix, see `fn parse_string(...)` in `lark/parser.rs`

PR #354, opened 2026-06-11
lark: support \xHH escapes in string literals

palios-taey:
Fixes the `\xHH` case of #54. Lark string literals are decoded through `serde_json`, which rejects `\xHH`, so this accepts `\xHH` in the string lexer and rewrites it to the equivalent `\u00HH` before decoding, with a test. Broader Python-Lark escape parity, notably `\UXXXXXXXX`, can follow separately.


palios-taey:
Heads up: the mypy failures here look unrelated to this change. It's `mypy 2.1.0` erroring on `numpy/__init__.pyi:737` (a PEP 695 `type` statement) under Python 3.14t — `Found 1 error in 1 file`, all in numpy's stub, nothing in the lark lexer/parser. This PR only touches Rust (`lexer.rs`, `parser.rs`, `test_lark.rs`). Looks like unpinned-mypy CI drift rather than anything in the diff. Happy to rebase whenever it's sorted.

hudson-ai:
Hi @palios-taey thank you for the contribution! Sorry for the long delay in reviewing. Looks great to me

PR #360, opened 2026-06-25
[Build] Deal with mypy/numpy issue

riedgar-ms:
The latest release of `numpy` (2.5) is shipping type hints which only work with Python 3.12 or later, but we still support earlier Python versions.  We configure `mypy` to make its compatibility check against 3.10, so `mypy` gets upset when running a later Python (which will have installed the latest `numpy`). This suppresses checking through `numpy` to avoid this.

### DIFF
--- a/python/mypy.ini
+++ b/python/mypy.ini
@@ -3,3 +3,11 @@ strict = True
 files = llguidance, torch_tests
 exclude = tmp
 python_version = 3.10
+
+# numpy >= 2.5 ships PEP 695 `type X = ...` aliases in its stubs, which mypy
+# only parses under python_version >= 3.12. We target 3.10 (requires-python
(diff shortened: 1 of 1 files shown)


hudson-ai:
Feels bizarre to have to declare this for a transitive dependency (actually -- maybe that's the underlying issue... if numpy is a dependency of some of our features, maybe we need to actually declare it in a feature group and be explicit about version bounds as a function of python version).

But looks good to me -- simply not following their definitions seems sane.

hudson-ai (approved):

riedgar-ms:
I agree that it's a bit of a `mypy` issue - why did it care how numpy annotated itself, we care about our code. Well, I suppose that it could just be a smaller portion of 'dependencies are hard' :-/

Issue #177, opened 2025-05-21
Preserve special tokens in `LLTokenizer.decode` method


hudson-ai:
Decoding a special token currently outputs `<[{token_id}]>` rather than the actual bytes corresponding to the token.

```python
class FakeTokenizer:
    def __init__(self):
        self.tokens = [bytes([i]) for i in range(256)] + [b"<s>"]
        self.special_token_ids = [256]
        self.bos_token_id = 256
        self.eos_token_id = 256

    def __call__(self, byte_string: bytes) -> list[int]:
        splits = byte_string.split(b"<s>")
        result = []
        for i, split in enumerate(splits):
            if i > 0:
                result.append(256)
            result.extend([b for b in split])
        return result

from llguidance import LLTokenizer, TokenizerWrapper

tok = LLTokenizer(
    TokenizerWrapper(
        FakeTokenizer()
    )
)

tokens = tok.tokenize_bytes(b"hello <s> world")
print(tokens)
print(tok.decode_bytes(tokens))
```
> [104, 101, 108, 108, 111, 32, 256, 32, 119, 111, 114, 108, 100]
> b'hello <[256]> world'

Expected: 
> [104, 101, 108, 108, 111, 32, 256, 32, 119, 111, 114, 108, 100]
> `b'hello <s> world'`

It seems this issue may only be present for tokenizers constructed via `TokenizerWrapper`? I cannot seem to repro with tokenizers constructed via `llguidance.hf.from_tokenizer`.

PR #363, opened 2026-07-14
Honor explicit special=false in ByteTokenizer::from_tokenizer

### LINKED ISSUE #361: from_tokenizer <...> heuristic overrides explicit special=false, dropping content tokens on decode
### Summary

`ByteTokenizer::from_tokenizer` classifies any added token shaped like `<...>` as special via a
name heuristic, even when the source `tokenizers` object explicitly flags that token
`special = false`. On `decode`/`decode_ext(..., include_special=false)` those tokens are then
stripped, which diverges from HuggingFace `transformers`
`tokenizer.decode(ids, skip_special_tokens=True)` (which honors the per-token flag and keeps them).

This is a narrow edge case for **content-carrying markup tokens** that happen to be `<...>`-shaped.

### Where

`toktrie_hf_tokenizers/src/lib.rs:188` (current `main`, crate v1.7.6); same heuristic at
`src/lib.rs:121-122` in v1.4.0 (the version resolved by downstream `mistral.rs`):

```rust
// we treat all added tokens of the form <...> as special tokens
if info.special || (info.content.starts_with("<") && info.content.ends_with(">")) {
```

I understand the `<...>` rule is intentional — it was added deliberately in #202 to make the
`llamacpp` path match the `transformers` path, since GGUF metadata often lacks a reliable special
flag. The gap is only in the second half of the `||`: when the tokenizer **does** provide an
explicit `special = false`, the heuristic still overrides it.

### Repro

A tokenizer with an added token flagged `special = false` whose content is `<...>`-shaped — e.g.
PaddleOCR-VL's OTSL table tokens `<fcel>` / `<nl>`, or Qwen2.5-VL's `<tool_call>`. These are
**content**, not control tokens:

```
transformers  tokenizer.decode([<fcel>, A], skip_special_tokens=True) -> "<fcel>A"   # honors special=false
tokenizers    (Rust crate)   .decode([<fcel>, A], true)               -> "<fcel>A"   # honors special=false
llguidance    tok_trie.decode_ext([<fcel>, A], include_special=false) -> "A"         # heuristic dropped <fcel>
```

For OCR this collapses a structured table (`<fcel>`/`<nl>` mark cells/rows) into a run-on string
with the structure gone. The token *ids* are fine — only the decode drops them.

### Proposed minimal change

Only apply the `<...>` shape heuristic when the tokenizer has **not** given an explicit special
flag for that token — i.e. let an explicit `special = false` win over the name shape. The
`info.special` branch and the GGUF/llamacpp fallback (where no reliable flag exists) stay exactly as
today; this only stops the shape heuristic from overriding a flag the tokenizer author set on
purpose.

Happy to send a PR if that direction sounds right, or adjust if you'd rather gate it behind an
option to preserve current behavior for callers that rely on it.

### Context

Surfaced while adding a document-OCR model downstream in `mistral.rs`; we've applied a local
correction there (honor the tokenizer's own `special` flag after `from_tokenizer`) and are filing
this so the root behavior can be considered upstream. Related prior art in this repo: #202 (origin
of the heuristic), #177 (preserve special tokens in `decode`).

sohumt123:
## Problem

`ByteTokenizer::from_tokenizer` classifies any added token shaped like `<...>` as a special token via a name heuristic, even when the source `tokenizers` object explicitly flags that token `special = false`:

```rust
// we treat all added tokens of the form <...> as special tokens
if info.special || (info.content.starts_with("<") && info.content.ends_with(">")) {
```

Special tokens get `TokTrie::SPECIAL_TOKEN_MARKER` (`0xff`) prepended, and `decode_ext(..., include_special=false)` then drops them. So **content-carrying** markup tokens that happen to be `<...>`-shaped — e.g. PaddleOCR-VL's OTSL table tags `<fcel>`/`<nl>`, or Qwen2.5-VL's `<tool_call>` — get silently stripped on a skip-special decode. That diverges from HuggingFace `transformers` and the Rust `tokenizers` crate, both of which honor the per-token flag:

```
transformers  decode([<fcel>, a], skip_special_tokens=True) -> "<fcel>a"   # honors special=false
tokenizers    decode([<fcel>, a], true)                     -> "<fcel>a"   # honors special=false
llguidance    decode_ext([<fcel>, a], include_special=false) -> "a"        # heuristic dropped <fcel>
```

The `<...>` shape rule was added deliberately in #202 to match the `transformers` path for GGUF, whose metadata often lacks a reliable special flag — and, per review, it also papers over tokenizers (e.g. Phi-3) that mislabel their control tokens, so other users may rely on it.

## Fix (updated per review)

Per the review feedback, the default behavior is now **unchanged**. This PR adds an opt-in constructor variant instead:

```rust
pub fn from_tokenizer_with_options(hft: Tokenizer, infer_special_tokens: bool) -> Result<ByteTokenizer>
```

- `infer_special_tokens = true` — the historical heuristic: any added token of the form `<...>` is treated as special regardless of its `special` flag.
- `infer_special_tokens = false` — only the tokenizer's explicit per-token `special` flag is honored, so content-carrying markup tokens like `<fcel>` survive a skip-special decode (the #361 case).

The existing `from_tokenizer` is re-implemented as a thin wrapper calling the new constructor with `infer_special_tokens = true`, so it is fully backward compatible. `from_file`/`from_json_bytes` (and the Python bindings, which go through them) keep the historical behavior as well.

## Testing

Two tests in the existing inline test module, both against a `tokenizer.json` with two `<...>`-shaped added tokens — `<fcel>` (`special = false`) and `<|im_end|>` (`special = true`):

- `from_tokenizer_with_options_honors_explicit_special_false` — with `infer_special_tokens = false`, the `special = false` token keeps its literal bytes (no `0xff` marker) and round-trips through `decode_ext([1, 0], false)` as `<fcel>a`; the `special = true` companion is still marked and stripped; name-based end-of-turn detection still fires (`tok_end_of_turn == Some(2)`); `include_special = true` keeps both.
- `from_tokenizer_default_infers_special_tokens` — the default constructor still marks **both** angle-bracket tokens special (old behavior preserved), strips both on skip-special decode, and keeps both with `include_special = true`.

```
running 16 tests
test tests::from_tokenizer_with_options_honors_explicit_special_false ... ok
test tests::from_tokenizer_default_infers_special_tokens ... ok
...
test result: ok. 16 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Also ran locally: `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings`, both clean. No FFI or Python-binding surface touched.

Related prior art: #202 (origin of the heuristic), #177 (preserve special tokens in decode).

Fixes #361

### DIFF
--- a/toktrie_hf_tokenizers/src/lib.rs
+++ b/toktrie_hf_tokenizers/src/lib.rs
@@ -72,7 +72,29 @@ impl ByteTokenizer {
     /// Creates a [`ByteTokenizer`] from an existing HuggingFace [`Tokenizer`],
     /// applying normalizer and pre-tokenizer fixes and extracting byte-level
     /// token representations.
-    pub fn from_tokenizer(mut hft: Tokenizer) -> Result<ByteTokenizer> {
+    ///
+    /// Added tokens shaped like `<...>` are treated as special tokens even if
+    /// the tokenizer flags them `special = false` (some tokenizers, e.g. Phi-3,
+    /// mislabel their control tokens). Use
+    /// [`ByteTokenizer::from_tokenizer_with_options`] with
+    /// `infer_special_tokens = false` to honor the explicit flag instead.
+    pub fn from_tokenizer(hft: Tokenizer) -> Result<ByteTokenizer> {
+        ByteTokenizer::from_tokenizer_with_options(hft, true)
+    }
+
+    /// Like [`ByteTokenizer::from_tokenizer`], but with explicit control over
+    /// special-token detection for added tokens.
+    ///
+    /// If `infer_special_tokens` is `true`, any added token of the form `<...>`
+    /// is treated as special regardless of its `special` flag (the historical
+    /// default; helps tokenizers that mislabel control tokens). If `false`, only
+    /// the tokenizer's explicit per-token `special` flag is honored, so
+    /// content-carrying markup tokens like `<fcel>` with `special = false` are
+    /// not stripped on decode (see issue #361).
+    pub fn from_tokenizer_with_options(
+        mut hft: Tokenizer,
+        infer_special_tokens: bool,
+    ) -> Result<ByteTokenizer> {
         let mut is_byte_level = false;
         let mut is_byte_fallback = false;
         let mut space_ch = ' ';
@@ -184,12 +206,14 @@ impl ByteTokenizer {
         let mut specials = HashSet::new();
 
         for (id, info) in added.iter() {
-            // Honor the tokenizer's explicit per-token `special` flag. The `<...>`
-            // shape heuristic from #202 only applies on paths that lack a reliable
-            // flag (e.g. GGUF/llamacpp); an explicit `special = false` from the
-            // `tokenizers` crate must win so content-carrying markup tokens are not
-            // dropped on decode. See #361.
-            if info.special {
+            // With `infer_special_tokens`, we treat all added tokens of the form
+            // <...> as special tokens; otherwise the tokenizer's explicit
+            // per-token `special` flag wins (see issue #361).
+            let is_special = info.special
+                || (infer_special_tokens
+                    && info.content.starts_with("<")
+                    && info.content.ends_with(">"));
+            if is_special {
                 match info.content.as_str() {
                     "</s>"
                     | "<|endoftext|>"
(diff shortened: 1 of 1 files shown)


hudson-ai:
@sohumt123 thanks for the contribution! I agree that this is indeed *correct* but I worry that (1) the original heuristic is actually useful for some "odd" tokenizers (I recall Phi 3 being in this group) and (2) that other users may be relying on this heuristic in some way.

I think I'd be more comfortable if you could add a constructor variant that takes a `infer_special_tokens` boolean and re-implement the existing constructor in terms of the new one with `infer_special_tokens=true`.

@mmoskal do you have an opinion here? I believe the heuristic originated from a commit you made a long time ago.

mmoskal:
I'm sure there were some tokenizers that had it wrong. Also not sure what exactly makes the token special, like why aren't the `<tool_call>` etc tokens special?

hudson-ai:
Thank you for the changes! Apologies for the slow feedback.

One more ask -- since you went for a generic name like `*_with_options`, let's future-proof it by providing options with a non-exhaustive struct like so:

```rust
#[derive(Debug, Clone, Default)]
#[non_exhaustive]
pub struct ByteTokenizerOptions {
    pub infer_special_tokens: bool,
}

pub fn from_tokenizer_with_options(
    hft: Tokenizer,
    options: ByteTokenizerOptions,
) -> Result<ByteTokenizer>
```

No other blockers from me :)

Issue #247, opened 2025-10-21
[Feature Request] Add generic whitespace regex support to avoid infinite whitespace issue


rajbirs-cerebras:
Currently, LLGuidance's regex separator support (implemented in PR #246) successfully handles flexible whitespace around `:` and `,` but lacks generic whitespace regex support for other JSON locations like after `{` and before `}`.

**Current Behavior**
✅ Supports regex whitespace patterns for separators (key_separator, item_separator) 
❌ Missing generic whitespace regex support for other JSON locations

**Problem**
Without generic whitespace regex support:
- No flexible control over whitespace after `{` and before `}`
- Limited to separator-specific whitespace patterns only
- Cannot prevent infinite whitespace in all JSON locations

**Requested Feature**
Add comprehensive generic whitespace regex support that allows flexible patterns throughout JSON, not just around separators.

**Use Case**
This would provide complete control over JSON whitespace generation using regex patterns, allowing bounded flexibility while preventing infinite whitespace generation across all JSON locations.

**Examples**
Current: Only controls separators
`{"key" :"value" , "key2"  : "value2"}  // ✅ Works`

Needed: Control all whitespace locations
```
{ "key": "value" }   // ✅ Should work with generic regex support
{   "key": "value"   } // ❌ Should be rejectable with bounds
```

**Benefits**

1. **Completeness**: Control whitespace in all JSON locations using regex patterns
2. **Consistency**: Extends the proven regex approach from PR #246
3. **Production Ready**: Prevents infinite whitespace while maintaining flexibility

**Related**
Original issue: #240
Separator regex PR: #246

behzad-a:
Can anyone look into this issue? cc @mmoskal @hudson-ai

hudson-ai:
Thanks for bumping @behzad-a. Looking into this on our side currently :)

okdshin:
Hi. I opened a PR (#367) that adds a `container_whitespace` option to bound whitespace around `{}`/`[]`, which should address this issue. [@hudson-ai](https://github.com/hudson-ai) could you take a look when you have a chance?

PR #367, opened 2026-07-24
[Feature] Add container_whitespace to bound whitespace around {}/[]

okdshin:
JSON structural whitespace directly after '{'/'[' and before '}'/']' is only produced by the repeatable lexer skip, so it can only be either 0 (whitespace_flexible=false) or unbounded (any skip pattern, since the skip lexeme repeats). That makes bounded-but-nonzero forms such as { "a": 1 } or [ 1, 2 ] impossible without risking runaway whitespace. The separators (item_separator/key_separator) already provide bounded, once-matched whitespace, but only around ':' and ','.

Add JsonCompileOptions.container_whitespace: Option<String>. When set, a once-matched whitespace lexeme (built like the separators) is inserted after '{'/'[' and before '}'/']', so a pattern such as [\x20\x0A\x0D\x09]{0,8} is enforced there. Defaults to None; settable via x-guidance.

Refs: guidance-ai/llguidance#247

### DIFF
--- a/parser/src/json/compiler.rs
+++ b/parser/src/json/compiler.rs
@@ -26,6 +26,7 @@ pub struct JsonCompileOptions {
     pub key_separator: String,
     pub whitespace_flexible: bool,
     pub whitespace_pattern: Option<String>,
+    pub container_whitespace: Option<String>,
     pub coerce_one_of: bool,
     pub lenient: bool,
     /// Allowed escape letters after '\' when quoting JSON strings.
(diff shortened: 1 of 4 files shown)


hudson-ai:
Hey @okdshin thank you for the contribution! 

Looks great to me, but I just opened PR #370 which should fix the more generic case -- do you still see a use-case for container-specific regex whitespace if that one is merged?

okdshin:
Thanks @hudson-ai! No, we don't have a use case that #370 wouldn't cover — a bounded `whitespace_pattern` gives us exactly what we needed here. Closing this in favor of #370.

PR #370, opened 2026-08-04
Add single-match skip lexemes

### LINKED ISSUE #247: [Feature Request] Add generic whitespace regex support to avoid infinite whitespace issue

hudson-ai:
## Summary

Add single-match behavior for skip lexemes: the lexemes generated from Lark `%ignore` directives and JSON `whitespace_pattern`. This prevents the skip regex from being reapplied indefinitely, so bounds such as `{1,8}` are enforced.

```lark
%llguidance { "ignore_once": true }
start: "A" "!"
%ignore /[ \t]{1,8}/
```

JSON `whitespace_pattern` automatically uses this behavior:

```json
{
  "x-guidance": {
    "whitespace_pattern": "[\\x20\\x0A\\x0D\\x09]{1,8}"
  }
}
```

## Motivation

Skip lexemes were previously accepted repeatedly. Consequently, patterns such as `\s` and `\s{1,8}` were effectively unbounded, making it difficult to constrain whitespace during LLM generation.

## Implementation

Skip configuration is represented internally by a `SkipSpec` containing the regex and its repetition behavior. Lark exposes the behavior as the boolean `%llguidance` option `ignore_once`.

After a once-mode skip is scanned, the parser copies the Earley row as before but removes that skip lexeme from the copied row's lexer state. Scanning an ordinary lexeme re-enables skip at the next grammar position.

This avoids grammar expansion, repetition counters, and per-token overhead. Nested grammar completion and `max_tokens` pops restore the appropriate parent skip configuration.

Existing `GrammarBuilder::add_grammar` and `LexerSpec::setup_lexeme_class` APIs remain available with their original unbounded behavior.

## Compatibility

Lark grammars remain unbounded by default; `ignore_once: true` is opt-in.

Default JSON behavior is unchanged:

- `whitespace_flexible: true` uses `[\x20\x0A\x0D\x09]+`.
- `whitespace_flexible: false` disables whitespace skip.

This is behavior-changing but effectively a bug fix for explicit `whitespace_pattern` values. Patterns such as `\s` or `\s{1,8}` are now matched at most once at each permitted whitespace location, so their upper bounds are enforced. Callers intentionally relying on repeated matching can express that repetition directly in the regex, for example `\s+`.

## Coverage

Tests cover:

- Legacy unbounded skip behavior
- Bounded Lark skip with `ignore_once: true`
- Bounded JSON whitespace
- Invalid `ignore_once` values
- Nested grammar skip restoration
- Parent skip restoration after `max_tokens`

Closes #247.

### DIFF
--- a/docs/json_schema.md
+++ b/docs/json_schema.md
@@ -79,10 +79,11 @@ Following keys are available inside of it:
 - `item_separator`, defaults to `","` - a regex pattern for the separator between array items or object properties
 - `key_separator`, defaults to `":"` - a regex pattern for the separator between object keys and values
 - `whitespace_flexible`, defaults to `true`; set to `false` to enforce compact JSON representation
-- `whitespace_pattern`, optional string, overrides `whitespace_flexible`;
-  `whitespace_flexible: true` is equivalent to `whitespace_pattern: r"[\x20\x0A\x0D\x09]+"`.
-  The pattern describes one complete whitespace span, so repetition bounds such as
+- `whitespace_pattern`, optional string, overrides `whitespace_flexible`. The pattern
+  describes one complete whitespace span, so repetition bounds such as
   `r"[\x20\x0A\x0D\x09]{1,8}"` are enforced. The span itself remains optional.
+  `whitespace_flexible: true` is equivalent to
+  `whitespace_pattern: r"[\x20\x0A\x0D\x09]+"`.
 - `coerce_one_of`, defaults to `false`; when set to `true`, the `"oneOf"` will be treated as `"anyOf"`
 - `json_allowed_escapes`, optional string, defaults to `"nrbtf\\\"u"`; controls which escape
   sequences are allowed after `\` in JSON strings. Each character in the string enables the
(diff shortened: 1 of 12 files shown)


riedgar-ms (approved):
I have no objection, but I'm not really familiar with all this....

hudson-ai:
@riedgar-ms I just changed the public api a bit: now using a boolean `ignore_once` option. Will merge if you have no objection to that :)

riedgar-ms:
No objections :-)

PR #372, opened 2026-08-05
Report empty integer intervals as unsatisfiable

### LINKED ISSUE #206: [Bug] Error on unsatisfiable integer schema not consistent

bryanwhl:
Fixes #206

Returns UnsatisfiableSchemaError when integer bounds contain no valid integer values, instead of exposing a regex-generation error.

### DIFF
--- a/parser/src/json/compiler.rs
+++ b/parser/src/json/compiler.rs
@@ -8,7 +8,7 @@ use indexmap::{IndexMap, IndexSet};
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
 
-use super::numeric::{check_number_bounds, rx_float_range, rx_int_range};
+use super::numeric::{check_number_bounds, normalize_integer_bounds, rx_float_range, rx_int_range};
 use super::schema::{build_schema, ArraySchema, ObjectSchema, OptSchemaExt, Schema};
 use super::shared_context::PatternPropertyCache;
 use super::RetrieveWrapper;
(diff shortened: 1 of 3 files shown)


parser/src/json/compiler.rs
```diff
@@ -297,6 +297,11 @@ impl Compiler {
             _ => None,
         }
         .map(|val| val as i64);
+        if matches!((minimum, maximum), (Some(min), Some(max)) if min > max) {
+            return Err(anyhow!(UnsatisfiableSchemaError {
+                message: "integer bounds contain no integer values".to_string(),
```
riedgar-ms: I don't think that this would catch the case where `min == max` but at least one is set to an exclusive bound?

bryanwhl: This case would already be caught by the earlier check_number_bounds() call, which rejects equal numbers based on their original bounds if either bounds are exclusive. Over here, we have normalized the values into inclusive integer endpoints (i.e. bounds), so `min == max` will always be valid here

riedgar-ms: Sorry, I hadn't noticed that call.

However, that brings up the question: why isn't this case being caught by `check_number_bounds()` ? It has
```rust
if min > max {
    return Err(format!(
        "{minimum_repr} ({min}) is greater than {maximum_repr} ({max})"
    ));
}
```
which ought to be equivalent, surely? I'm guessing (skimming through the code and usage) that it's due to `exclusiveMax = exclusiveMin + 1` is fine for `number` but not for `int`.

This is a useful improvement to the error message, but I think it belongs in `check_number_bounds()` (which might in turn merit splitting into a couple of sub-functions, given that it also checks `multipleOf`). I see that `NumberSchema` does have an `integer` flag, so `check_number_bounds()` should have all the information it needs.

bryanwhl: @riedgar-ms Do you mind checking my updated implementation? I think this way of abstraction looks right, so that we reuse `normalize_integer_bounds()` consistently across the 2 places that needs it.

riedgar-ms (approved):
Thanks!

parser/src/json/numeric.rs
```diff
@@ -506,6 +506,42 @@ pub fn rx_float_range(
...
+}
+
+fn check_integer_bounds(num: &NumberSchema) -> Result<(), String> {
+    let (minimum, maximum) = normalize_integer_bounds(num);
+    if matches!((minimum, maximum), (Some(min), Some(max)) if min > max) {
+        return Err("integer bounds contain no integer values".to_string());
+    }
+    Ok(())
+}
+
```
hudson-ai: I know the caller is already a pretty long function, but removing this indirection would allow a more useful error message, e.g. "integer interval (0, .9] contains no integer values".

bryanwhl:
@hudson-ai Pinging for a review here

hudson-ai:
Apologies for the delay! Looks great. Thank you :)

PR #374, opened 2026-08-10
json: allow general Unicode escapes in strings and object keys

mmoskal:
## Summary

- Add the opt-in `json_allow_general_unicode_escapes` compiler option for unconstrained JSON string values and object keys.
- Accept printable Unicode escapes such as `\u00e9`, require valid UTF-16 surrogate pairs, and count each decoded Unicode scalar once for `minLength` and `maxLength`.
- Apply `json_allowed_escapes` consistently to additional-property keys and remove the separate permissive key regex.
- Cache compiled Unicode string regexes to avoid repeated parsing and lexer-fuel exhaustion.
- Update three MaskBench snapshots because additional-property keys now reject invalid surrogate escapes and honor `json_allowed_escapes`; their validation outcomes are unchanged.
- Keep regex-constrained strings and constants on their existing matching paths, and expose the new option through the Python bindings and JSON Schema documentation.

## Why

The existing compiler rejects valid JSON strings such as `"\u00e9"` because printable characters are expected as literal UTF-8. Allowing general Unicode escapes requires validating surrogate pairs and preserving JSON Schema string-length semantics.

## Verification

- `cargo +stable test -p llguidance --quiet` (1,533 tests passed)
- `cargo +stable check --workspace`
- `cargo +stable clippy -p llguidance -p llguidance_py --all-targets -- -D warnings`
- Full MaskBench corpus: all 11,306 expected results match


parser/src/json/compiler.rs
```diff
@@ -335,7 +352,11 @@ impl Compiler {
 
     fn json_simple_string(&mut self) -> NodeRef {
         cache!(self.string_cache, {
-            let ast = self.json_quote(RegexAst::Regex("(?s:.*)".to_string()));
+            let ast = if self.options.json_allow_general_unicode_escapes {
+                self.json_general_unicode_string(0, None).unwrap()
```
hudson-ai: Could you make this function return `Result<NodeRef>` and propagate the `Result` rather than unwrapping here? I don't \~think\~ that will impact too many call sites

parser/src/json/compiler.rs
```diff
@@ -823,6 +848,75 @@ impl Compiler {
...
+        }
+
+        let scalar = if let Some(expr) = self.general_unicode_scalar_cache {
+            expr
+        } else {
+            let allowed_escapes = self
+                .options
+                .json_allowed_escapes
+                .as_deref()
+                .unwrap_or("nrbtf\\\"u");
```
hudson-ai: `nrbtf\\\"u`  is now duplicated between  `json_quote()`  and  `json_general_unicode_string()`  (and mirrors the default used by  derivre), so these paths could drift. Would you factor the default into a shared constant/helper?

parser/src/json/compiler.rs
```diff
@@ -32,6 +32,9 @@ pub struct JsonCompileOptions {
     /// Defaults to full JSON set: nrbtf"u\
     /// For example, set to nrbtf"\ to disallow \uXXXX escapes.
     pub json_allowed_escapes: Option<String>,
+    /// Allow printable Unicode escapes in strings and object keys without regex constraints.
+    /// Paired surrogate escapes count as one character for string length limits.
+    pub json_allow_general_unicode_escapes: bool,
```
hudson-ai, line 37: I was planning on doing a minor release. This is technically a breaking change if we're strict about semver, but I think the risk is low here. @mmoskal , @riedgar-ms opinions? Might be a good idea to mark the struct as non-exhaustive, although I believe that's an even wider-reaching breaking change...

mmoskal: Do you mean that we don't allow invalid surrogates in key names? Or the additional field?

parser/src/json/compiler.rs
mmoskal: fixed

parser/src/json/compiler.rs
mmoskal: fixed; simpler without the macro

parser/src/json/compiler.rs
hudson-ai: The additional field. I'm not worried about key names, but adding the public field to a public struct is *technically* breaking

mmoskal: Sorry, didn't read properly. `#[non_exhaustive]` would also break `JsonCompileOptions { lenient: true, ..Default::default() }` which is sort of silly.

We seem to follow Python not Rust effects on semver: 1.6.0 added json_allowed_escapes here. I think this is reasonable, I suspect there is many more Python than Rust users, and having separate versions doesn't seem worth it.

hudson-ai (approved):
LGTM, thanks!

PR #377, opened 2026-08-21
fix: handle non-lossless forced tokenization

### LINKED ISSUE #366: LLMatcher.consume_token() rejects a token fill_next_token_bitmask() just marked as the only legal choice (multi-byte key_separator + merged quote-colon token)
**Environment:** llguidance 1.7.6, transformers 5.14.1, macOS/Apple Silicon (but this is tokenizer/grammar logic, not platform-specific).

**Summary:** `LLMatcher.consume_token()` rejects a token that `fill_next_token_bitmask()` had just reported, at the same matcher state, as the *only* legal next token. This happens with a JSON-schema grammar whose `x-guidance.key_separator` is a multi-byte string (`": "`) against the `stepfun-ai/Step-3.7-Flash` tokenizer, which has a merged `"":"` (quote+colon) token in its vocabulary.

**Repro (self-contained, tokenizer-only download, no model weights):**

```python
import json
import llguidance
import llguidance.hf
import llguidance.numpy
import numpy as np
from llguidance import LLMatcher
from transformers import AutoTokenizer

hf_tok = AutoTokenizer.from_pretrained("stepfun-ai/Step-3.7-Flash", trust_remote_code=True)
llg_tok = llguidance.hf.from_tokenizer(hf_tok)

grammar = json.dumps({"grammars": [{"name": "json_schema", "json_schema": {
    "type": "object",
    "properties": {"title": {"type": "string", "minLength": 1}},
    "required": ["title"],
    "x-guidance": {"item_separator": ", ", "key_separator": ": ",
                   "whitespace_flexible": False, "whitespace_pattern": "",
                   "coerce_one_of": False, "lenient": False,
                   "json_allowed_escapes": None},
}, "lark_grammar": None}], "max_tokens": None})

m = LLMatcher(llg_tok, grammar)
mask = llguidance.numpy.allocate_token_bitmask(1, llg_tok.vocab_size)

text = ""
for step in range(6):
    llguidance.numpy.fill_next_token_bitmask(m, mask, 0)
    bits = np.unpackbits(mask[0].view(np.uint8), bitorder="little")
    allowed = np.flatnonzero(bits)
    tid = int(allowed[0])
    m.consume_token(tid)
    text += hf_tok.decode([tid])
    err = m.get_error()
    print(step, repr(hf_tok.decode([tid])), "n_allowed_at_this_step=", allowed.size, "error=", err)
    if err:
        break
```

**Output:**

```
0 '{"' n_allowed_at_this_step= ...
1 'title' n_allowed_at_this_step= ...
2 ':' n_allowed_at_this_step= 1 error= Parser Error: token ":" doesn't satisfy the grammar; forced bytes: got '"'; applying ':'
```

(Full attached script has one more variant, and a version with `key_separator=":"` -- no trailing space -- which does *not* fail on this tokenizer.)

The tokenizer has a single token that decodes to `":` (id 28 in the log below is the bare `:`; a separate id decodes to the merged `":`). At step 2 the bitmask returned by `fill_next_token_bitmask` marks exactly one token as legal, the caller consumes exactly that token, and `consume_token` still reports an error and enters forced-bytes recovery mid-consumption. Mask generation and token consumption appear to be evaluating the grammar's expectation of the (multi-byte) `key_separator` differently.

**What does *not* reproduce it:**
- The same schema with `key_separator` set to a single-character `":"` (no trailing space) walks the grammar to completion without error on the same tokenizer.
- The same `key_separator: ": "` schema against a different tokenizer (tested: a Gemma-4 MoE tokenizer) also walks to completion without error.

So the trigger appears to be specifically: multi-byte `key_separator` (or plausibly any multi-byte forced string) combined with a tokenizer vocabulary that has a token spanning across the separator's byte boundary (here, a token that itself contains `":` -- quote followed by colon -- so the matcher's forced-bytes state machine and the token-level mask disagree about where inside the separator that merged token leaves the parser).

**Where we hit this:** LM Studio's structured-output path (via `mlx_vlm`'s `LLGuidanceLogitsProcessor`) uses exactly this shape of schema (a `{"title": string}` grammar for auto-generating chat titles) against this exact model, and it crashes the generation thread every time a title-generation request runs. Filing here since the model-loading/inference stack around it was ruled out first (reproduces identically with zero model weights loaded, using only `LLMatcher` directly).

Happy to attach the full script as a file if that's more convenient than inline.

yu-xin-c:
## Summary

- detect when fast-forward tokenization normalizes forced bytes and cannot round-trip them
- fall back to byte-prefix masking so healing offsets stay aligned with parser state
- add a tokenizer-agnostic regression test covering normalization across a JSON key separator

## Root cause

`ff_tokens()` computed healing offsets from decoded token lengths even when tokenization had normalized the original forced byte sequence. With the Step-3.7 tokenizer, the separator space in `": "` was removed during tokenization, so mask generation allowed a bare colon that token consumption rejected.

## Validation

- `cargo test -p llguidance`
- `cargo clippy -p llguidance --all-targets --all-features -- -D warnings`
- `cargo check --no-default-features -p llguidance`
- `cargo fmt --all -- --check`
- reproduced with `stepfun-ai/Step-3.7-Flash`: bare colon is excluded and every allowed token at the failing state consumes successfully

Fixes #366

### DIFF
--- a/parser/src/tokenparser.rs
+++ b/parser/src/tokenparser.rs
@@ -707,9 +707,18 @@ impl TokenParser {
         if do_force {
             let t0 = Instant::now();
             let (mut tokens, mut num_fixed) = self.token_env.tokenize_bytes_marker(&forced_bytes);
+            let trie = self.token_env.tok_trie();
+            // Healing offsets are invalid when tokenization normalizes the forced bytes.
+            if num_fixed == 0 && trie.decode_raw(&tokens) != forced_bytes {
+                infoln!(
+                    self,
+                    "non-lossless forced-byte tokenization; falling back to byte prefix"
+                );
+                token_prefix = forced_bytes[num_existing_bytes..].to_vec();
+                return (Vec::new(), token_prefix);
+            }
             if !tokens.starts_with(&existing_tokens) {
                 // whoops, re-tokenize without the prefix
-                let trie = self.token_env.tok_trie();
                 infoln!(
                     self,
                     "re-tokenizing without prefix: {}; because we got {}",
--- a/parser/tests/test_raw_parser.rs
+++ b/parser/tests/test_raw_parser.rs
@@ -2,7 +2,7 @@ use lazy_static::lazy_static;
 use llguidance::{
     api::TopLevelGrammar,
     earley::SlicedBiasComputer,
-    toktrie::{ApproximateTokEnv, InferenceCapabilities, TokEnv, TokenizerEnv},
+    toktrie::{ApproximateTokEnv, InferenceCapabilities, TokEnv, TokRxInfo, TokTrie, TokenizerEnv},
     Matcher, ParserFactory, TokenParser,
 };
 use serde_json::{json, Value};
@@ -61,6 +61,42 @@ fn consume(parser: &mut TokenParser, tok: u32) {
     assert!(n == 0);
 }
 
+struct NormalizingTokEnv {
+    trie: TokTrie,
+}
+
+impl NormalizingTokEnv {
+    fn new() -> Self {
+        let mut words = (0..=255).map(|byte| vec![byte]).collect::<Vec<_>>();
+        words.extend([
+            br#"{""#.to_vec(),
+            b"title".to_vec(),
+            br#"":"#.to_vec(),
+            br#"":""#.to_vec(),
+            b"\xff<eos>".to_vec(),
+        ]);
+        let info = TokRxInfo::new(words.len() as u32, words.len() as u32 - 1);
+        Self {
+            trie: TokTrie::from(&info, &words),
+        }
+    }
+}
+
+impl TokenizerEnv for NormalizingTokEnv {
+    fn tok_trie(&self) -> &TokTrie {
+        &self.trie
+    }
+
+    fn tokenize_bytes(&self, bytes: &[u8]) -> Vec<u32> {
+        let normalized = bytes
+            .iter()
+            .copied()
+            .filter(|byte| *byte != b' ')
+            .collect::<Vec<_>>();
+        self.trie.greedy_tokenize(&normalized)
+    }
+}
+
 #[test]
 fn test_ff_tokens() {
     let lark = r#"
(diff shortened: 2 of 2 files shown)


parser/src/tokenparser.rs
```diff
@@ -707,9 +707,18 @@ impl TokenParser {
         if do_force {
             let t0 = Instant::now();
             let (mut tokens, mut num_fixed) = self.token_env.tokenize_bytes_marker(&forced_bytes);
```
mmoskal: [P2] Check the second tokenization for normalization too

This round-trip check runs before the retry without the previous token at lines 728–730. With a tokenizer that strips leading spaces and a vocabulary containing a merged `a b` token, tokenizing `a b` is lossless but does not preserve the existing `a` token. The retry then tokenizes ` b` into `b`, bypassing the new check.

I reproduced this with `start: ("a" | "c") " b"`: after consuming `a`, the mask allows only `b`, but consuming it fails because the grammar still requires the space. Please validate the retry's output against its input before computing healing offsets, and cover this path in the regression test.

parser/src/tokenparser.rs
```diff
@@ -707,9 +707,18 @@ impl TokenParser {
...
+            let trie = self.token_env.tok_trie();
+            // Healing offsets are invalid when tokenization normalizes the forced bytes.
+            if num_fixed == 0 && trie.decode_raw(&tokens) != forced_bytes {
```
mmoskal: [P2] Fixed tokens bypass validation of surrounding text

`num_fixed > 0` means the sequence contains an explicit or special token; it does not guarantee that ordinary text following that token round-trips. This condition skips the normalization check for that text as well.

Using a tokenizer that strips leading spaces, I reproduced this with `start: <[257]> " b"`, where token 257 is a special token. After consuming it, the mask allows `b`, but consuming `b` fails because the required space was normalized away. Please validate the ordinary text segments while retaining explicit-token semantics, and add coverage for required text after a special token.

mmoskal (changes requested):
Two cases still allow the token mask to disagree with token consumption when forced text is normalized. Both were reproduced with a small in-memory tokenizer against this PR's head.

parser/src/tokenparser.rs
yu-xin-c: Fixed in abad0ec. The retry input is now checked independently before healing offsets are computed. The regression test uses `start: ("a" | "c") " b"` with a leading-space-normalizing tokenizer and verifies that bare `b` is rejected while the required space is accepted.

parser/src/tokenparser.rs
yu-xin-c: Fixed in abad0ec. Validation now walks marker-aware segments and checks each ordinary text segment without treating explicit markers as text. The new regression covers `<[SPECIAL]> " b"`, verifies the explicit token is still fast-forwarded, and then checks that bare `b` is rejected while the required space is accepted.

yu-xin-c:
Both requested cases are addressed in abad0ec: retry tokenization is validated independently, and ordinary text is validated segment-by-segment around explicit token markers while preserving safe fast-forwarded tokens. The original Step-3.7 reproduction and all three in-memory normalization regressions pass, along with the full llguidance test suite and clippy. The new workflow runs currently require maintainer approval; a re-review when convenient would be appreciated.

PR #380, opened 2026-09-02
GPU-Resident Fwd-Projected Masking via PDA Export

sempervictus:
# Pushdown Automata Export for Context-Free Grammars

## What this does

Adds a `dpda` feature to llguidance that compiles any context-free grammar
(CFG) to a **pushdown automaton (PDA)** and exports a flat, GPU-uploadable
transition table (the `CudaPackage`).

This is the CPU-side export. The downstream GPU runtime (separate PR)
consumes the table to enforce grammar constraints on token sampling and
speculative drafting without a CPU round-trip.

## Why this is needed

A **pushdown automaton** adds a bounded stack to the finite control. For
**deterministic** CFGs (DCFLs - which includes all practical JSON schemas,
tool grammars, and regex-derived constraints), the stack depth is bounded
by the longest production (typically 4-8 for JSON/tool grammars), making
the PDA a finite table.

## What's in this PR

- `dpda` feature flag (gates the pushdown-rs dependency)
- `dpda_adapter.rs`: the `PdaGrammar` wrapper that implements the
  `pushdown_rs::Grammar` trait for `CGrammar`. Precomputes global-to-local
  symbol index maps for O(1) lookups during RTN compilation.
- `compile_pda(&CGrammar) -> PdaMachine`: the RTN construction
  (Alpay & Senturk, arXiv:2603.05540, Definition 5) with exact
  `kappa(G)` state count.
- `export_pda_package(&CGrammar) -> CudaPackage`: the flat POD bitvec +
  CSR u32 offsets + source primitives. This is the H2D payload.
- 9 tests in `test_dpda.rs` (using llguidance's own single-byte test
  tokenizer, 256 vocab):
  - JSON schema compiles to a valid PDA (structure + bounds)
  - Regex [a-z]+ compiles to a valid PDA (deterministic)
  - Bitvec round-trips losslessly
  - CudaPackage exports valid POD (header matches)
  - kappa(G) matches the compiled state count
  - Differential: PDA accepts == independent CFG oracle accepts
  - JSON PDA rejects empty input
  - Regex PDA is deterministic + rejects empty
  - Start mask is non-empty
- 2 docs: `pda_theoretical_foundation.md` (6 theorems with proofs and
  oracles) + `pda_gpu_integration.md` (the runtime contract)
- README: "GPU Offload (PDA)" section

## The CudaPackage layout

```
Header (7 u32s):
  num_states, num_inputs, num_stack_syms, num_transitions,
  num_accepting, start_state, start_stack

Accepting (num_accepting u32s):
  The accepting state IDs

Transitions (variable-length records):
  Per record: q, a, top, next_q, push_len, push[0..push_len-1]
  Total array length = sum over all records of (5 + push_len) u32s

CSR u32 offsets (num_states u32s):
  ctrl_u32_offsets[q] = the u32 index into the transition array where
  state q's records begin. O(1) jump for the GPU kernel.
```

## Limits

- **DCFL only.** Ambiguous grammars compile to a non-deterministic PDA.
  The `is_deterministic()` check gates this: if the grammar is ambiguous,
  the export fails and the CPU fallback is used.
- **Bounded stack.** The stack depth D is fixed at compile time (default 8).
  Grammars with nesting deeper than D will reject.
- **No parametric grammars.** Rules with runtime parameters (the `perm::_`
  style) cannot be flattened to a PDA. The adapter returns an error.

## Reproducibility

```bash
cargo test -p llguidance --features dpda --test test_dpda   # 9/9
```

The tests use llguidance's own single-byte test tokenizer (256 vocab,
`ApproximateTokEnv::single_byte_env()`). No external fixtures required.

## References

- Alpay & Senturk, "Attention Meets Reachability: Structural Equivalence and
  Efficiency in Grammar-Constrained LLM Decoding,"
  [arXiv:2603.05540](https://arxiv.org/abs/2603.05540) (2026).
  The RTN compilation (Def 5), kappa(G) state count (Def 10, Lemma 2).
- Collura et al., "Stay Within Your Bounds: Distance-Guided Decoding for
  Guaranteed Context-Free Grammar Compliance,"
  [arXiv:2608.28229](https://arxiv.org/abs/2608.28229) (2026).
  The SWYB bounded pushdown summary.
- Li et al., "Efficient Grammar-Constrained Decoding via Parser Stack
  Classification," [arXiv:2608.03065](https://arxiv.org/abs/2608.03065) (2026).
  The PSC mask-classification.
- Chen et al., "Pre3: Enabling Deterministic Pushdown Automata for Faster
  Structured LLM Generation,"
  [arXiv:2506.03887](https://arxiv.org/abs/2506.03887), ACL 2025.
  The LR(1)->DPDA + prefix-conditioned edges (the fusion advantage).
- Park et al., "Flexible and Efficient Grammar-Constrained Decoding,"
  [arXiv:2502.05111](https://arxiv.org/abs/2502.05111) (2025).
  The token spanner (T_inv) + the stack invariance.
- Hopcroft, Motwani, Ullman, "Introduction to Automata Theory, Languages, and
  Computation" (3rd ed.). The PDA/DPDA definitions.
- Sipser, "Introduction to the Theory of Computation." The DCFL = the
  DPDA-recognizable languages.

## Credit

The algorithms implemented here are proven results from the cited arXiv papers,
years of osmotic learning from work alongside one of the original RISC team
members from IBM Research (Norman Schibuk), and brutally guided (self-hosted)
LLM co-implementation (over a ~5M token context - YARN works); not an invention.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -187,6 +187,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1b8e56985ec62d17e9c1001dc89c88ecd7dc08e47eba5ec7c29c7b5eeecde967"
 
 [[package]]
+name = "bitvec"
+version = "1.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ddcec3d12c579d40898fe0a9a358a803c23e9c52ca3c425707f81c9436211837"
+dependencies = [
+ "funty",
+ "radium",
+ "tap",
+ "wyz",
+]
+
+[[package]]
 name = "borrow-or-share"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -747,6 +759,12 @@ dependencies = [
  "lazy_static",
  "num",
 ]
+
+[[package]]
+name = "funty"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e6d5a32815ae3f33302d95fdcb2ce17862f8c65363dcfd29360480ba1001fc9c"
 
 [[package]]
 name = "futures"
@@ -1405,6 +1423,7 @@ dependencies = [
  "jsonschema",
  "lazy_static",
  "llg_test_utils",
+ "pushdown-rs",
  "rand",
  "rayon",
  "referencing",
@@ -1862,6 +1881,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "pushdown-rs"
+version = "0.1.0"
+source = "git+https://github.com/sempervictus/pushdown-rs#da365889af7522aa3573264a437228e2222a6b2c"
+dependencies = [
+ "bitvec",
+ "rten-simd",
+]
+
+[[package]]
 name = "pyo3"
 version = "0.29.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1933,6 +1961,12 @@ name = "r-efi"
 version = "5.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "69cdb34c158ceb288df11e18b4bd39de994f6657d83847bdffdbd7f346754b0f"
+
+[[package]]
+name = "radium"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dc33ff2d4973d518d823d61aa239014831e521c75da58e3df4840d3f47749d09"
 
 [[package]]
 name = "rand"
@@ -2170,6 +2204,12 @@ dependencies = [
  "syn",
  "unicode-ident",
 ]
+
+[[package]]
+name = "rten-simd"
+version = "0.26.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7f0c01294eb782adca256aa130b48a04006d03763073a68f45a9c33b982ea556"
 
 [[package]]
 name = "rustc-demangle"
@@ -2543,6 +2583,12 @@ dependencies = [
  "core-foundation-sys",
  "libc",
 ]
+
+[[package]]
+name = "tap"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "55937e1799185b12863d447f42597ed69d9928686b8d88a1df17376a097d8369"
 
 [[package]]
 name = "target-lexicon"
@@ -3413,6 +3459,15 @@ name = "writeable"
 version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ea2f10b9bb0928dfb1b42b65e1f9e36f7f54dbdf08457afefb38afcdec4fa2bb"
+
+[[package]]
+name = "wyz"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "05f360fc0b24296329c78fda852a1e9ae82de9cf7b27dae4b7f62f118f77b9ed"
+dependencies = [
+ "tap",
+]
 
 [[package]]
 name = "yoke"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1883,7 +1883,7 @@ dependencies = [
 [[package]]
 name = "pushdown-rs"
 version = "0.1.0"
-source = "git+https://github.com/sempervictus/pushdown-rs#da365889af7522aa3573264a437228e2222a6b2c"
+source = "git+https://github.com/sempervictus/pushdown-rs#ceb709c24a93322b2aaa0ad1cc9499ccb4c082ff"
 dependencies = [
  "bitvec",
  "rten-simd",
--- a/README.md
+++ b/README.md
@@ -81,6 +81,23 @@ The library implements a context-free grammar parser using Earley’s algorithm 
 
 Grammars can be also used to speed up decode via [fast-forward tokens](./docs/fast_forward.md).
 
+### GPU Offload (PDA)
+
+For inference runtimes with GPU sampling (e.g. xinfer, vLLM, TensorRT-LLM),
+llguidance can export a **pushdown automaton (PDA)** table that runs entirely
+on the GPU, eliminating the CPU round-trip for grammar masking.
+
+- **Per-token sampling:** the PDA mask is fused into the GPU sampling kernel
+  (one launch: mask + sample + advance). No CPU involvement during generation.
+- **Drafting (MTP/DFlash):** the PDA projects K+1 masks ahead in one kernel
+  launch, constraining the speculative draft on-device.
+- **Optional dispatch:** if no grammar is active, the PDA path is skipped
+  entirely (zero overhead). The existing unmasked sampling/drafting runs.
+
+Enable with `--features dpda`. See [docs/pda_gpu_integration.md](./docs/pda_gpu_integration.md)
+for the full integration guide and [docs/pda_theoretical_foundation.md](./docs/pda_theoretical_foundation.md)
+for the correctness proofs.
+
 ### Comparison and performance
 
 See [MaskBench](https://github.com/guidance-ai/jsonschemabench/tree/main/maskbench) in
--- a/docs/pda_gpu_integration.md
+++ b/docs/pda_gpu_integration.md
@@ -1,0 +1,198 @@
+# PDA GPU Integration Guide
+
+**For:** the GPU inference runtime integrating llguidance's grammar PDA.
+**Depends on:** `pushdown-rs` (the PDA engine), `attention-rs` (the CUDA kernels).
+
+## 1. Model Load (CPU, once)
+
+```rust
+use llguidance::dpda_adapter::{compile_pda, export_pda_package};
+use llguidance::api::TopLevelGrammar;
+use llguidance::ParserFactory;
+
+// 1. Build the grammar (Lark, JSON schema, regex, or explicit token ranges).
+let g = TopLevelGrammar::from_lark(grammar_source);
+
+// 2. Compile to a CGrammar (the Earley parser's internal form).
+let mut parser = factory.create_parser(g)?;
+parser.start_without_prompt();
+let grm = parser.parser.grammar().clone();
+
+// 3. Compile to a PDA (the RTN construction, kappa(G) states).
+let pda = compile_pda(&grm)?;
+// pda.num_states, pda.transitions, pda.accepting, pda.start_state, pda.start_stack
+
+// 4. Export the CUDA package (the flat POD bitvec + source primitives).
+let pkg = export_pda_package(&grm)?;
+// pkg.bitvec: BitVec<u64> (the GPU-uploadable payload)
+// pkg.num_states, pkg.num_inputs, pkg.num_stack_syms, pkg.num_transitions
+// pkg.upload_bytes(): the H2D size (~210 KB for the tool-call grammar)
+
+// 5. Upload to GPU (attention-rs).
+let table = PdaPushdownTable::from_cuda_package(&pkg, device)?;
+// table.transitions: the flat (q, a, top, next_q, push_len, push[]) array
+// table.accepting: the accepting state IDs
+// table.num_states, table.num_inputs, table.num_stack_syms, table.num_transitions
+```
+
+## 2. Per-Token Sampling (GPU, fused)
+
+The fused_sample kernel performs the full per-token PDA step in one launch:
+scan the transition table for the current (ctrl, top), emit the VOB mask,
+apply it to the logits, sample, and advance the PDA state.
+
+**The CSR index.** The transition table is a flat array of variable-length
+records. Without an index, each GPU thread would scan all 8764 records to
+find the ~80 that match its control state. The CSR (ctrl_u32_offsets,
+ctrl_counts) lets each thread jump directly to its state's records in O(1).
+The CSR are computed on the CPU at upload time (one pass over the
+transition array) and uploaded as a small [num_states] u32 tensor.
+
+**The logit format.** The PDA mask is applied to F32 or BF16 logits. The
+model's lm_head produces logits in the model's compute dtype; the sampler
+dequantizes to F32 before the PDA step. FP8/FP4 are weight formats (for
+the GEMM), not logit formats - by the time logits reach the sampler they
+are already F32/BF16. The PDA kernel does not need FP8/FP4 variants.
+
+**The optional dispatch.** If `ctrl` is null (no grammar active), the kernel
+runs plain sampling with zero PDA overhead. The PDA is strictly optional.
+
+```rust
+// The fused kernel: ONE launch does mask + sample + advance.
+// If no grammar is active, pass ctrl=null (the kernel skips the PDA path).
+
+let (out_ctrl, out_sp, out_tokens) = table.fused_sample(
+    &logits,      // [batch, vocab] F32/BF16
+    &ctrl,        // [batch] current control state (or null)
+    &stack,       // [batch * D] bounded stack
+    &sp,          // [batch] stack pointer
+    &PdaSampling::TopKTopP { temperature: 0.8, top_k: 50, top_p: 0.9 },
+)?;
+// out_ctrl: [batch] next control state
+// out_sp: [batch] next stack pointer
+// out_tokens: [batch] sampled token IDs
+```
+
+### The Optional Dispatch
+
+```rust
+// No grammar: zero PDA overhead (the existing sampling path).
+let tokens = sampler.sample_cuda_vob(&logits, k, p, temp, seed, None)?;
+
+// Grammar active: the PDA mask is fused on-GPU (fused into sampling).
+let (ctrl, sp, tokens) = pda_table.fused_sample(&logits, &ctrl, &stack, &sp, &sampling)?;
+```
+
+## 3. Drafting (MTP/DFlash, GPU, fused)
+
+The projection kernel walks K draft tokens through the PDA and emits K+1
+VOB masks in one launch. K is a runtime parameter (the adaptive-k system
+changes it per step based on acceptance rate), so the kernel is launched
+individually rather than captured in a CUDA graph. The graph is only used
+for the fixed model forward pass (attention + MLP + lm_head).
+
+```rust
+// The projection kernel: ONE launch emits K+1 VOB masks.
+// Used to constrain a speculative draft: each draft position gets the
+// mask of the PDA state reached after the preceding draft tokens.
+
+let projected = table.fused_project(
+    &ctrl,        // [batch] current control state
+    &stack,       // [batch * D] bounded stack
+    &sp,          // [batch] stack pointer
+    &draft,       // [batch, K] draft tokens
+)?;
+// projected: [batch * (K+1) * words_per_vob] U32 VOB masks
+
+// Convert to the DFlash allow matrix format: [batch*(K+1), vocab] F32
+let allow = table.vob_to_allow(&projected, batch, k, vocab)?;
+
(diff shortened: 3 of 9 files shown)


sempervictus:
# PDA Theoretical Foundation

## 0. Intent

Provide inference runtimes with **inline grammar masking on the GPU itself**
for context-free grammars, for two loops:

1. **Regular sampling (line-rate grammar enforcement).** Every generated token is advanced
   through the grammar PDA and its mask applied to the logits, with no CPU
   round-trip. This enforces control-flow integrity: out-of-order control
   tokens are masked out at the position where they are illegal.

2. **Speculative / MTP / DFlash drafting.** The draft model proposes K tokens;
   the GPU advances the PDA K positions ahead, validates each against the
   grammar, and produces the per-position masks - all on-device. K is
   caller-supplied (capped by the SWYB `d_H` bound, default 16).

The CPU (llguidance) is responsible for *building* the PDA tables once; the
GPU is responsible for *running* them per token. State crosses the CPU/GPU
boundary as a compact `(ctrl, stack, sp)` tuple, so neither side recomputes
the other's work.

## 1. Definitions

Let G be a context-free grammar (CFG) over a tokenizer T with vocabulary V
(|V| = n). A **configuration** is a pair `(q, gamma)` where `q` is a control
state and `gamma` is the stack contents. The **mask** at a
`(q, gamma)` is the set of input symbols `a` for which a transition
`(q, a, top(gamma)) -> (q', push)` exists.

The **RTN compilation** (Alpay & Senturk, Def 5) maps G to a PDA A_G with:
- Control states Q = {q_start} union {q_A^in, q_A^out : A in N} union {q_(p,i) : p in P, i in 0..=|rhs(p)|}
- Stack alphabet Gamma = {bot} union {q_(p,i) : p in P, i} (the return addresses)
- Transitions: start, choice, terminal, call, return, exit (Def 5)

The exact state count is **kappa(G) = 1 + 2|N| + sum_p(|rhs(p)| + 1)**
(Def 10, Lemma 2).

## 2. Theorems and Proofs

### Thm 1 (Compilation correctness)

For every CFG G, the RTN-compiled PDA A_G accepts exactly L(G).

**Proof.** By construction (Def 5): the start transition pushes the
bottom marker; the choice transitions select a production; the terminal
transitions shift; the call transitions push a return address; the return
transitions pop and goto; the exit transition reaches the accepting state.
The stack discipline ensures that a nonterminal call returns to the correct
dot position. By induction on the derivation tree, every string in L(G)
has an accepting computation, and every accepting computation yields a
string in L(G). QED.

**Oracle:** `pushdown_rs::oracle::cfg_accepts` (the naive recursive CFG
definition, zero shared code with the PDA). The differential test
(`test_dpda.rs::differential_cfg_pda_oracle`) verifies 100% agreement
on a corpus of all inputs up to length 6.

### Thm 2 (Determinism for DCFL grammars)

If G is a deterministic CFG (no two productions share the same lhs with
overlapping first sets), the RTN-compiled PDA is deterministic (at most
one transition per `(q, a, top)`).

**Proof.** The RTN construction creates one choice transition per
production per stack symbol. If the grammar is deterministic, at most one
production is applicable at any position, so at most one choice transition
fires. The terminal and call transitions are keyed by `(q, a, top)`
which is unique by construction. QED.

**Oracle:** `PdaMachine::is_deterministic()` (sorts all `(q, a, top)`
keys, checks for duplicates). The test `anb_n_is_deterministic` verifies
the {a^n b^n} DPDA is deterministic.

### Thm 3 (Bounded stack for non-recursive CFGs)

For a CFG G with bounded recursion depth R (the longest chain of nonterminal
calls in any derivation), the PDA stack depth is bounded by D = R + 1.

**Proof.** Each nonterminal call pushes one return address onto the stack.
The maximum nesting depth is the length of the longest nonterminal call
chain, which is R by definition. The bottom marker adds 1. Total: D = R + 1.
For JSON schemas and tool-call grammars, R is typically 4-8 (the maximum
nesting depth of the schema),), so D = 5-9. The GPU kernel
allocates D u32 registers per thread. QED.

**Note.** For recursive grammars (e.g. S -> a S b | eps), R is unbounded
(the recursion depth grows with the input length). Such grammars produce
PDAs with unbounded stacks, which are NOT suitable for GPU execution.
The `max_stack_depth` parameter (default 8) acts as a hard limit: if the
stack exceeds D during a computation, the PDA rejects. This is correct for
bounded-recursion grammars (the stack never exceeds D) but means that
unbounded-recursion grammars will reject sufficiently long inputs.

**Oracle:** `PdaMachine::accepts_dpda` (the DPDA simulation tracks the
stack; if it exceeds D, the computation is rejected). The test
`proof_stack_is_bounded` verifies the push length per transition is <= 2
for the {a^n b^n} DPDA (the maximum push in a single step, not the total
stack depth over the entire computation).

### Thm 4 (Bitvec losslessness)

The `to_bitvec` / `from_bitvec` round-trip is the identity: for any
PdaMachine M, `from_bitvec(to_bitvec(M)) == M`.

**Proof.** The bitvec layout is a flat sequence of u32 fields:
header (7 u32s) + accepting IDs + per-transition records
(q, a, top, next_q, push_len, push[0..push_len-1]). The `from_bitvec`
deserializer reads exactly this layout and reconstructs the PdaMachine.
The `validate_bounds` check ensures all references/inputs/stack symbols are
in range. Since the serialization is a bijection between PdaMachine and
valid bitvecs, the round-trip is lossless. QED.

**Oracle:** `test_dpda.rs::bitvec_round_trips` and
`pda_tests.rs::proof_bitvec_roundtrip_is_identity`.

### Thm 5 (Projection equals sequential)

The K-step projection `project_batch(configs, drafts)` produces the same
masks as K sequential `step_batch` calls.

**Proof.** By induction on K: the base case (K=0) is the mask at the
initial config. The inductive step: the mask at position i+1 is computed
from the config after advancing by draft[i], which is exactly what the
sequential step does. The projection unrolls this loop. QED.

**Oracle:** `pda_tests.rs::proof_projection_equals_sequential` and
`proof_project_batch_simd` (the batch invariant: SIMD projection ==
scalar projection).

### Thm 6 (Layout-identity: GPU == SIMD == scalar)

The CUDA kernel, the SIMD mock, and the scalar reference all consume the
same bitvec layout. The GPU result is correct iff the scalar result is
correct (Thm 1) and the kernel performs the documented lookup (Thms 3-5).

**Proof.** The `CudaPackage` is the bitvec + source primitives. The GPU
kernel reads the transition table in the same flat layout
`(q, a, top, next_q, push_len, push[])`. The lookup algorithm is
identical: scan for matching `(ctrl, top)`, collect allowed inputs into
the VOB. The only difference is the execution substrate (GPU thread vs
CPU lane vs scalar loop). Since the algorithm is the same and the table
is the same, the results are the same. QED.

**Oracle:** The GPU kernel is validated by comparing the same bitvec
through the CPU `PdaMachine::accepts` and comparing. The GPU unit tests
(`pda_fused_sample_matches_cpu_reference`, `pda_fused_project_masks`)
verify the layout-identity invariant on a real GPU.

## 3. Verification Table

| Theorem | Statement | Oracle | Test |
|---------|-----------|--------|------|
| Thm 1 | PDA accepts L(G) | `cfg_accepts` (independent) | `differential_cfg_pda_oracle` |
| Thm 2 | DCFL => deterministic | `is_deterministic()` | `anb_n_is_deterministic` |
| Thm 3 | Stack bounded by D | `accepts_dpda` (rejects > D) | `proof_stack_is_bounded` |
| Thm 4 | Bitvec lossless | round-trip == identity | `bitvec_round_trips` |
| Thm 5 | Projection == sequential | batch invariant | `proof_projection_equals_sequential` |
| Thm 6 | GPU == SIMD == scalar | layout-identity | `proof_cuda_graph_dag_and_replay` |

## 4. Experimental Results

| Grammar | Type | States | Transitions | GPU mem | Export time |
|---------|------|--------|-------------|---------|-------------|
| Tool-call envelope | Lark + token ranges | 107 | 8,764 | ~210 KB | <1 s |
| JSON 3-field object | JSON schema | 106 | 7,613 | ~180 KB | <1 s |
| [a-z]+ | regex | 7 | 14 | ~2 KB | <1 s |
| a^n b^n (recursive) | Lark | 13 | 48 | ~1 KB | <1 s |
| select(5) | Lark | 19 | 271 | ~5 KB | <1 s |

All grammars compile to valid PDAs with `kappa(G)` matching the state count.
The tool-call envelope grammar is the largest realistic case:
107 states, 8764 transitions, 210 KB GPU upload.

## 5. Completeness and Limits

1. **DCFL only.** The PDA is deterministic (Thm 2). Ambiguous grammars
(multiple derivations for the same string) compile to an NPDA, which
    the GPU kernel does not support (the scan finds multiple transitions
   for the same `(q, a, top)`). The `is_deterministic()` check gates this.

2. **Bounded stack.** The stack depth D is fixed at compile time (the
   `max_stack_depth` parameter, default 8). Grammars with nesting deeper
than D will reject (the PDA rejects when the stack exceeds D). For
    JSON and tool grammars, D=8 covers all realistic nesting.

3. **Full-vocab mask.** The VOB mask is `ceil(num_inputs/32)` u32 words.
   For a 248K vocab, this is 7750 words = 31 KB per state. The GPU kernel
   uses a CSR index (ctrl_u32_offsets) to jump directly to the relevant
   transitions for the current control state (O(1-3) records per state,
   not O(all transitions)). For the tool-call grammar (8764 transitions,
   107 states), each state carries ~80 transitions on average, but the
   CSR limits the scan to only the matching state's records.

4. **No parametric grammars.** Rules with runtime parameters (the
   `perm::_` style) cannot be flattened to a PDA. The adapter returns
   an error for these (the `validate()` check).

5. **GPU latency (measured on RTX 5090, sm_120, batch=32, vocab=4096):**
   - fused_sample: 157 us/step (one kernel launch: mask + sample + advance)
   - fused_project (K=16): 65 us/step (one kernel launch: K+1 masks)
   - CPU baseline (reactive Earley): 17-238 us per token
   - The GPU PDA is memory-bound at large vocabs (248K: ~33 GiB/s, the
     L3/main bandwidth limit). At 4K vocab the compute is the bottleneck.

## 6. Reproducibility

```bash
# The PDA engine (31 core proofs + 6 SIMD accuracy proofs):
cargo test -p pushdown-rs --features simd

# The SIMD bench (scalar vs SIMD at 4 vocab sizes):
cargo bench -p pushdown-rs --bench simd_bench

# The llguidance adapter (9 proofs):
cargo test -p llguidance --features dpda --test test_dpda

# The real-tokenizer example (248K vocab, tool-call + 8 lark grammars):
cargo run -p llguidance --features dpda --example dpda_real_tokenizer

# The GPU kernel validation (requires a CUDA host; see the runtime's test suite):
# The layout-identity invariant (Thm 6) is verified by the GPU unit tests
# in the runtime that consumes the CudaPackage.
```

sempervictus:
# PDA GPU Integration Guide

**For:** the GPU inference runtime integrating llguidance's grammar PDA.
**Depends on:** `pushdown-rs` (the PDA engine), `attention-rs` (the CUDA kernels).

## 1. Model Load (CPU, once)

```rust
use llguidance::dpda_adapter::{compile_pda, export_pda_package};
use llguidance::api::TopLevelGrammar;
use llguidance::ParserFactory;

// 1. Build the grammar (Lark, JSON schema, regex, or explicit token ranges).
let g = TopLevelGrammar::from_lark(grammar_source);

// 2. Compile to a CGrammar (the Earley parser's internal form).
let mut parser = factory.create_parser(g)?;
parser.start_without_prompt();
let grm = parser.parser.grammar().clone();

// 3. Compile to a PDA (the RTN construction, kappa(G) states).
let pda = compile_pda(&grm)?;
// pda.num_states, pda.transitions, pda.accepting, pda.start_state, pda.start_stack

// 4. Export the CUDA package (the flat POD bitvec + source primitives).
let pkg = export_pda_package(&grm)?;
// pkg.bitvec: BitVec<u64> (the GPU-uploadable payload)
// pkg.num_states, pkg.num_inputs, pkg.num_stack_syms, pkg.num_transitions
// pkg.upload_bytes(): the H2D size (~210 KB for the tool-call grammar)

// 5. Upload to GPU (attention-rs).
let table = PdaPushdownTable::from_cuda_package(&pkg, device)?;
// table.transitions: the flat (q, a, top, next_q, push_len, push[]) array
// table.accepting: the accepting state IDs
// table.num_states, table.num_inputs, table.num_stack_syms, table.num_transitions
```

## 2. Per-Token Sampling (GPU, fused)

The fused_sample kernel performs the full per-token PDA step in one launch:
scan the transition table for the current (ctrl, top), emit the VOB mask,
apply it to the logits, sample, and advance the PDA state.

**The CSR index.** The transition table is a flat array of variable-length
records. Without an index, each GPU thread would scan all 8764 records to
find the ~80 that match its control state. The CSR (ctrl_u32_offsets,
ctrl_counts) lets each thread jump directly to its state's records in O(1).
The CSR are computed on the CPU at upload time (one pass over the
transition array) and uploaded as a small [num_states] u32 tensor.

**The logit format.** The PDA mask is applied to F32 or BF16 logits. The
model's lm_head produces logits in the model's compute dtype; the sampler
dequantizes to F32 before the PDA step. FP8/FP4 are weight formats (for
the GEMM), not logit formats - by the time logits reach the sampler they
are already F32/BF16. The PDA kernel does not need FP8/FP4 variants.

**The optional dispatch.** If `ctrl` is null (no grammar active), the kernel
runs plain sampling with zero PDA overhead. The PDA is strictly optional.

```rust
// The fused kernel: ONE launch does mask + sample + advance.
// If no grammar is active, pass ctrl=null (the kernel skips the PDA path).

let (out_ctrl, out_sp, out_tokens) = table.fused_sample(
    &logits,      // [batch, vocab] F32/BF16
    &ctrl,        // [batch] current control state (or null)
    &stack,       // [batch * D] bounded stack
    &sp,          // [batch] stack pointer
    &PdaSampling::TopKTopP { temperature: 0.8, top_k: 50, top_p: 0.9 },
)?;
// out_ctrl: [batch] next control state
// out_sp: [batch] next stack pointer
// out_tokens: [batch] sampled token IDs
```

### The Optional Dispatch

```rust
// No grammar: zero PDA overhead (the existing sampling path).
let tokens = sampler.sample_cuda_vob(&logits, k, p, temp, seed, None)?;

// Grammar active: the PDA mask is fused on-GPU (fused into sampling).
let (ctrl, sp, tokens) = pda_table.fused_sample(&logits, &ctrl, &stack, &sp, &sampling)?;
```

## 3. Drafting (MTP/DFlash, GPU, fused)

The projection kernel walks K draft tokens through the PDA and emits K+1
VOB masks in one launch. K is a runtime parameter (the adaptive-k system
changes it per step based on acceptance rate), so the kernel is launched
individually rather than captured in a CUDA graph. The graph is only used
for the fixed model forward pass (attention + MLP + lm_head).

```rust
// The projection kernel: ONE launch emits K+1 VOB masks.
// Used to constrain a speculative draft: each draft position gets the
// mask of the PDA state reached after the preceding draft tokens.

let projected = table.fused_project(
    &ctrl,        // [batch] current control state
    &stack,       // [batch * D] bounded stack
    &sp,          // [batch] stack pointer
    &draft,       // [batch, K] draft tokens
)?;
// projected: [batch * (K+1) * words_per_vob] U32 VOB masks

// Convert to the DFlash allow matrix format: [batch*(K+1), vocab] F32
let allow = table.vob_to_allow(&projected, batch, k, vocab)?;

// Feed into the DFlash candidate walk (the existing masked path).
let selected = dflash_select_candidates_masked(
    &hidden, &unary_logits, &candidate_ids,
    &predecessor_codebook, &successor_codebook, &anchor_token,
    Some(&allow),  // None for unmasked DFlash
)?;
```

### The MTP Verify Path

```rust
// MTP verify: the target model scores K+1 positions in parallel.
// The PDA projection constrains each position's logits.
// If no grammar is active, the PDA projection is skipped (zero overhead).

let pda_masks: Option<&Tensor> = match pda_table {
    Some(table) => Some(&table.fused_project(&ctrl, &stack, &sp, &draft)?),
    None => None,  // No grammar: plain MTP verify
};
```

## 4. CPU<->GPU State Transfer

```
Model load:
  CPU: export_pda_package -> CudaPackage (bitvec + primitives)
  GPU: PdaPushdownTable::from_cuda_package (H2D upload, once)

Per token:
  GPU: fused_sample -> (out_ctrl, out_sp, out_tokens)
  CPU: (no round-trip needed; the state stays on GPU)

Return boundary (end of generation):
  GPU: D2H (out_ctrl, out_sp) -> CPU
  CPU: pda_validate_draft (the CPU-side check, if needed)
  CPU: sync_from_pda (re-align the Earley parser, O(delta))
```

The CPU advantage over the old DFA approach: the PDA state is
`(ctrl, stack[D], sp)` - a fixed-size tuple that fits in GPU registers.
No CPU round-trip is needed during generation. The state only crosses
the boundary at the return point (end of sequence).

## 5. The Layout-Identity Invariant

```
  pushdown-rs bitvec (CPU)
       |
       |  to_pod_bytes()
       v
  CudaPackage (the flat POD)
       |
       |  H2D upload
       v
  PdaPushdownTable (GPU)
       |
       |  fused_sample / fused_project
       v
  GPU results
       |
       |  D2H + compare to CPU PdaMachine::accepts
       v
  Layout-identity proof: GPU == CPU (same table, same algorithm)
```

The CPU `PdaMachine` is the correctness oracle for the GPU kernel.
If they disagree, the kernel is wrong (the table is proven correct by
the pushdown-rs test suite, 30/30).

sempervictus:
In case anyone (or anything, presumably) reviewing this wonders why the integration notes re CUDA are explicitly here, this is what we're seeing in pretty naive benchmarks on laptop kit for a fused DFA kernel:
```
  HW: Intel Ultra 9 290HX Plus + RTX 5090 Laptop (SM120)
  Accuracy: validate_draft GPU=16 CPU=16 match=true
            project_masks GPU == expected match=true

  sample_and_advance (fused kernel, vocab=4, K=16, 10000 iters):
    batch   us/launch   us/seq    vs CPU parser (50us)
    1       9.57        9.57      5x
    4       9.29        2.32      22x
    8       9.52        1.19      42x
    16      9.43        0.59      85x
    32      9.51        0.30     168x
```
with a bit of plumbing this will hopefully allow much less latency/overhead for the constrained generation of tokens and increase adoption both for its utility and security properties when covering the full envelope of generation. Rather difficult to force a model to emit an out of place injected BOS when you have
```lark
start: reasoning_block ( text | tool_call )+ eos
reasoning_block: (<[^248044-248046,248058-248059,248068-248069]>)+ (<[248069]>)
text: (<[^248044-248046,248058-248059,248068-248069]>)+
tool_call: <[248058]> tool_content <[248059]>
param_0_0: "\n<parameter=label>\n" value_string
param_0_1: "\n<parameter=task>\n" value_string
tool_0: "\n<function=spawn>"(param_0_0)?  param_0_1 "</function>\n"
...
```
masking it from the whole generation space.

sempervictus:
Pardon, didn't select the correct drop-down when i opened it.
Working out integration details with various grammar types (hard anchors, repetitions, etc) for prac-ap to avoid having to generate a couple hundred thousand states when only a few will do.

sempervictus:
Nested JSON schemas and complex grammars with close boundaries dont seem to be happy with even a high entropy DFA keying off Parser rows/states. Eventually the time tradeoff to project all potential future states for transitions becomes too significant. PDA's might be the better abstraction to pull projected states into the GPU - investigating. This actually still works for the vast majority of input under a negation mask grammar (prevent explicit tokens/combinations over a range) but we'd  rather construct a mechanism that can handle both the majority and minority of cases since fallback to CPU for those edges is technically "a lot more expensive" (no more so than before for those particular tokens) but more importantly creates an irregularity in what we hope to use as a security boundary upon which to build.

sempervictus:
Re-architected the whole thing - too many structural variants and states for DFA to express effectively or efficiently. Implemented a crate for various DPAs to support this and other RISC-y functionality, currently validating in staging environment through `attention-rs` and `xinfer` integration.

PR #382, opened 2026-09-09
fix: cancel abandoned matcher computations

### LINKED ISSUE #261: Asynchronous Cancellation for Speculative Decoding
Consider a low latency LLM service with an arbitrary draft model topology for speculative decoding, arbitrary acceptance rates, beam search, and multiple requests being served concurrently. Consider also arbitrarily or even intentionally complex grammars coming from external users of the service. Constrained decoding will use numerous Matchers in parallel to accurately explore multiple paths, consuming considerable cpu resources. Many failed speculative paths will be pruned, leaving runaway tasks on the cpu until mask generation completes.

To save cpu resources, it is desirable to abandon compute intensive mask generation operations that are in progress for pruned paths. With tuning, parser limits provide a mechanism to mitigate the issue, and works well to prevent long mask ops, but it is not suitable for short circuiting a failed speculation, since setting it too low may be punitive.

The request is for an asynchronous cancellation mechanism to cancel compute intensive operations of the Matcher, with support for C bindings. One possible implementation... Early return can be done the same way as when parser limits are reached, and the parser limits object may also be used to pipe through a shared flag that may be set asynchronously from outside the library (perhaps under a convenient `cancel` api). The Matcher may be left in an indeterminant but safe to destroy state post-cancellation, without resource leaks. This may also be extended to the StopController if appropriate.

Thank you,
Chris

RanaPriyansh:
Abandoned speculative paths continue to consume CPU while matcher computations finish.

Add safe matcher cancellation and make its runtime checks opt-in. Rust callers use `into_cancellable()`. The C cancellation-handle getter enables cancellation before it returns the handle.

Local verification passed 1,655 Rust tests in debug and release, 121 C/C++ cases with 4,645 assertions, and 21 Python matcher tests. Python validation used a development extension because macOS rejected the release extension. Formatting, Clippy, feature, build, and generated-header checks passed.

Four balanced benchmark rounds show residual, host-specific overhead. With cancellation disabled, the three mask-size medians were 3.5%, 5.7%, and 7.2% above the exact base. Other cases were noisy. The benchmark reply contains all selected workloads.

Fixes #261.

AI assistance: Codex was used as a collaborator.


hudson-ai:
Hi @RanaPriyansh thank you for the contribution! I think that cancellation is definitely a worthwhile feature, both to support speculative decoding and possibly to support mask computations with timeouts for certain high performance use cases.

Since the measured steady-state overhead is 5–13%, I think that it would be worthwhile to explore letting cancellation be opt-in per matcher.

For example, the parser could hold `Option<CancellationHandle>` (defaulting to  `None`), and expose an  `into_cancellable()` operation. The disabled path would retain only a predictable option check and avoid atomic loads, while callers that need asynchronous cancellation would explicitly accept the polling cost.

If the remaining option checks still have measurable overhead, cancellable and non-cancellable traversal could potentially be specialized later (but an optional handle seems like a relatively small first step). Would you be able to benchmark this approach against the current implementation?

RanaPriyansh:
Thanks. I tested the optional-handle design and updated the branch. New Rust matchers leave cancellation disabled until `into_cancellable()` is used. The C handle getter enables cancellation before it returns the handle.

I ran four balanced Criterion rounds on an Apple M3. Each run used 30 samples, a 0.5-second warm-up, and a one-second measurement period.

| Case | Disabled/base | Disabled/current PR | Enabled/disabled |
| --- | ---: | ---: | ---: |
| `compute_mask/8192` | +3.5% | +3.6% | +3.5% |
| `compute_mask/32768` | +5.7% | -7.6% | +5.4% |
| `compute_mask/128000` | +7.2% | -3.5% | +2.4% |
| `compute_mask_positions/after_key` | -0.4% | -7.8% | -1.2% |
| `compute_mask_positions/in_string` | +2.2% | -8.7% | +4.3% |
| `token_generation/32768` | +1.3% | -6.5% | -2.5% |
| `first_mask/32768` | -3.2% | -4.3% | +0.5% |
| `lazy_lexeme/greedy_charclass` | -6.3% | -9.5% | +16.3% |

`first_mask` includes construction and opt-in allocation. Several cases had large outliers. These results are descriptive host measurements, not a general overhead bound.

Local verification passed 1,655 Rust tests in debug and release, 121 C/C++ cases, and 21 Python matcher tests. Python validation used a development extension because macOS rejected the release extension.

hudson-ai:
Hey @RanaPriyansh, thanks for testing out the opt-in version!

I'm still a bit concerned with the performance impact on the non-cancellable branch, but I did some poking around and I believe that it's solvable. I think that the overall approach should be to move cancellation polling out of the per-byte trie traversal hot path, e.g. by checking once whether cancellation is enabled in the recognizer and specializing  add_bias_inner  to an enabled and disabled variant. The enabled case can poll between bounded chunks of work rather than checking whether to poll on every byte down in  try_push_byte . A chunk size of around 1024 nodes retained roughly 8–15µs cancellation response in testing, so your current polling interval of 16 has quite a bit of wiggle room.

RanaPriyansh:
Implemented the chunked traversal approach in `510f6b6d05a892112120f7891b0fa13712e75691`. `add_bias_inner` selects enabled or disabled variants once, and enabled traversal polls between chunks of 1,024 flattened trie positions. The per-byte countdown is gone. Extension traversal also stops between chunks.

I also deferred cancellation-handle inspection until uncached lexer work. ARM64 assembly inspection confirms that the disabled cached byte path avoids the previous option load and test. Nested lexer and parser checks remain, along with speculative-state cleanup and final mask rejection.

I compared exact upstream `bae21e50`, previous PR head `9253518e`, and the updated code on Apple M3 with Rust 1.95.0. All modes used the same benchmark workload and lockfile. Cancellation-enabled and disabled runs used the same executable for each revision.

The table covers three in-string synthetic vocabulary cases with default slicing. I ran eight paired rounds with rotating mode order. Each Criterion run used 100 samples, three seconds of warm-up, and five seconds of measurement. Values are medians of round-level ratios of Criterion mean estimates. All rounds, including large outliers, are retained.

| Vocabulary | Updated disabled / upstream | One-sided 95% upper bound | Updated disabled / previous disabled | Updated enabled / previous enabled |
|---|---:|---:|---:|---:|
| 8,192 | +0.75% | +3.31% | −6.71% | −8.22% |
| 32,768 | +0.23% | +0.50% | −8.90% | −10.31% |
| 128,000 | +0.53% | +4.86% | −10.83% | −11.68% |

The upper bounds resample whole paired rounds. These results show a median improvement over the previous PR, but do not establish overhead below 2% for every case. An incomplete same-binary control used an earlier candidate, so it does not establish the final binary's noise level. I am not claiming zero overhead or attributing every timing swing to host noise.

Full masks matched across all five revision/cancellation modes with slicing enabled and disabled for these three cases and a pinned Phi tokenizer case. The canonical setup used real token IDs and verified that slicing actually ran. Those checks establish output equivalence for these fixtures, not a full tokenizer performance survey.

A separate instrumented 1,024-chunk run sampled 128 request offsets, spaced 16 attempts apart across a 2,048-attempt phase. Request-to-worker-return latency was 5.08 µs median, 10.04 µs p95, 13.42 µs p99, and 27.75 µs maximum. Sampled extra `try_push_byte` calls after the request had median 516 and maximum 1,020. Timing includes the cancellation store, barrier release, and worker scheduling. These are sample measurements, not a fixed response guarantee.

The final source passed 1,664 Rust tests in both debug and release, plus Clippy, formatting, no-default-feature, build, generated-header, and sample-parser checks. Relinked C/C++ and ASan wrapper tests passed. ASan instrumentation covered the C/C++ wrapper. The rebuilt Python development extension passed 21 matcher tests. Python release imports previously failed on both the candidate and exact upstream with the same macOS `LINKEDIT` error. Full external Python and platform coverage remains a CI requirement.

hudson-ai:
Awesome, thanks! That's looking a lot better in my opinion. Thanks for doing all of this!

I have one remaining concern though. Changing `compute_ff_tokens()`, `consume_ff_tokens()`, and `compute_ff_bytes()` to return `Result` breaks the existing public API. These methods already represent parser errors by returning an empty result and leaving the matcher in an error state, so I think that cancellation can work the same way. We can preserve the existing API and let callers inspect `is_error()`, `is_cancelled()`, `get_error()`, or `stop_reason()` afterward. The C binding can still return -1 by checking the matcher state after the call.

I think that `StopReason::Cancelled` is useful, although adding a variant to the public enum is also technically a breaking change for downstream exhaustive matches. Not huuugely concerned about this one though.

RanaPriyansh:
Restored the original vector return types in `39aab5620c5ac023e3f08a947fe5da4801b6b7c0` for `compute_ff_tokens()`, `consume_ff_tokens()`, and `compute_ff_bytes()`. Cancellation returns empty output and leaves the matcher in its terminal error state. Callers can inspect `is_error()`, `is_cancelled()`, `get_error()`, or `stop_reason()`.

The C forced-token function checks matcher state before copying output. It returns -1 on failure and preserves the output buffer. Python forced-output methods also retain their original empty-output behavior.

The API review found two additional signature changes in the public `RegexVec` methods. Their original arguments are restored. Cancellation-specific implementations remain crate-private. `StopReason::Cancelled` remains available.

The same external signature test fails against `510f6b6` with five type errors and passes against the corrected source. Tests cover cancellation during forcing and consumption, terminal error state, and preserved error causes.

Local validation passed 1,668 Rust tests in debug and release, with zero skipped. Formatting, Clippy, no-default-feature checks, required builds, generated-header validation, and eight sample-parser runs passed. Relinked C/C++ and ASan wrapper tests passed. The rebuilt Python development extension passed 22 matcher tests.

The release run had one [nextest leaky annotation](https://www.nexte.st/docs/features/leaky-tests/). An isolated repeat passed without that annotation. Its cause is not established. Full external Python and platform CI remain pending.

PR #383, opened 2026-09-12
PDA (pushdown-automaton) grammar-constrained decoding for llguidance

sempervictus:
# PDA (pushdown-automaton) grammar-constrained decoding for llguidance

This PR brings the bounded PDA (pushdown automaton) state machine into the
llguidance decode path, replacing the unbounded Earley item-set mask with an
O(kappa) PDA mask that is GPU-resident and constant in sequence length. 

## What changed

- `dpda_adapter.rs` (new): the `PdaGrammar` bridge (implements the
  `pushdown_rs::Grammar` trait over the llguidance `CGrammar`), `compile_pda`
  (CFG -> PDA via the RTN construction), `export_pda_package` (the `CudaPackage`
   H2D payload), and `terminal_token_map` / `terminal_token_map_dfa` (the
   terminal -> vocabulary-token bridge, the finite token spanner).
- `earley/parser.rs`: the PDA machine is compiled once at `ParserState::new`
  and advanced in lockstep with the Earley parser (the P11 sync invariant).
  `compute_bias` gains a PDA mask path (the O(kappa) replacement for the
  O(item_set) Earley mask); `validate_tokens` gains a PDA projection path.
- `tokenparser.rs` / `matcher.rs`: the non-advancing `_immut` mask/ff reads
  and the safe snapshot (a speculative-decoding gate keeps the sequence and
  the FSM in lockstep without perturbing the read state).
- `pda_config` / `pda_machine` / `pda_forced_token` exposed on the `Matcher`
  for downstream consumers (the xinfer runner, the attention-rs fused kernels).

## The maths (from `pushdown-rs/docs`)

**RTN compilation** (`mathematics.md`): a CFG `G = (N, Sigma, P, S)` compiles
to a PDA `M_G` whose control states are
`Q = {q_start} U {q_A^in, q_A^out : A in N} U {q_(p,i) : p in P, i in 0..=|rhs(p)|}`.
The exact control-state count is

    kappa(G) = 1 + 2|N| + sum_p (|rhs(p)| + 1)

(the Lemma 2 of arXiv:2603.05540). The stack holds return addresses (the dot
states); bounded by `D = max_p |rhs(p)| + 1`. The language is preserved:
`L(M_G) = L(G)` (a string is accepted by the PDA iff it is derived by the CFG).

**The six constrained properties** (`finite_stateless_grammar.md`) that make a
grammar compilable to a finite, uploadable, SIMD-executable DPDA table:
finite, Stateless, Bounded control (kappa exact), Deterministic (DCFL, the
Valiant 1973 result), Bounded pushdown (fixed-size stack), Finite token
spanner (the precomputed bridge). A grammar violating any of the six stays on
the CPU (the general NPDA / the Earley parser).

**Epsilon-closure advance** (`mathematics.md`): the RTN choice/call/exit/return
moves are epsilon, so a terminal step must first resolve the epsilon closure.
`advance_eps(q, stk, a)` BFSes the epsilon moves (bounded by `MAX_EPS_CLOSURE`)
to the config where the terminal move `a` is available, then takes it. The CSR
index (`ctrl_offsets` + `ctrl_counts`, computed at construction) makes the
per-state transition lookup O(1-3) instead of O(total). The GPU kernel uses the
same CSR format, so the CPU and GPU paths are consistent (the P8 property).

## Advantages

- **Bounded state**: the PDA mask is O(kappa) (constant in sequence length)
  vs the Earley mask O(item_set) (grows with length, hits the 50000-item limit
  on recursive grammars). This is the 60 -> 10 tokens/sec fix on long messages.
- **GPU-resident**: the PDA table (the bitvec POD) uploads once at model load;
   the per-step mask/advance/project run on the GPU (the attention-rs fused
   kernels), not the CPU.
- **Deterministic single path**: for a DCFL the epsilon closure is a single path
   (no BFS blowup), which is the SIMD-friendly case.
- **Same table, three tiers**: the scalar, the SIMD, and the GPU all consume the
   same bitvec (the layout-identity invariant), so the GPU port is proven by the
   CPU oracle without a GPU.

## Proofs / correctness

The pushdown-rs test suite (the `pda_tests`, the `proof_*`) establishes:
- **Language correctness**: `L(M_G) = L(G)` (the {a^n b^n} accepts/rejects, the
   differential vs an independent CFG-membership oracle, exhaustive + boundary corpora).
- **Determinism**: `is_deterministic` (the at-most-one transition per
  (q, a, top)); the DPDA single path vs the NPDA bounded BFS.
- **Bounded stack**: the reachable depth <= D; the per-step push bound.
- **Mask fidelity**: the PDA mask == the machine's legal inputs over the full
   config space (the inclusive + exclusive oracle).
- **Projection**: `project(K) == K sequential steps` (the break on divergence).
- **Epsilon-closure consistency**: `mask_at_cfg == the advance_eps-able inputs`
  over the reachable config space (the `proof_mask_batch_consistent_with_advance_eps`).
- **Bitvec round-trip**: `to_bitvec -> from_bitvec` is lossless (the exact POD
  size; the `PdaMachine` equality excludes the derived CSR fields and the
  CPU-side `vocab_names`, so the POD identity holds across tiers).
- **Three-way identity**: scalar == SIMD batch == GPU kernel (the same bitvec;
  the SIMD mock is the GPU's oracle, the scalar is the SIMD's oracle).

The llguidance side adds the `test_dpda` battery (15 proofs: structure, kappa,
bitvec round-trip, CudaPackage export, JSON/regex/nested-JSON language checks,
start-mask non-empty, differential vs an independent oracle, accepts a valid
string) and the `test_pda_differential` / `test_pda_sequence_growth` suites.

## Qualifications / limits

- **DCFL only**: ambiguous grammars fail the `is_deterministic` check and stay
  on the CPU (the NPDA / the Earley path).
- **Bounded stack**: `D = max |rhs| + 1`; recursive grammars with unbounded
  nesting reject long inputs (the `max_stack` safeguard).
- **No parametric grammars**: the PDA is compiled from the concrete CFG; a
   parametric (the runtime-parameterized) grammar has no static table.
- **The PDA mask is a superset gate**: for non-deterministic grammars the
  epsilon-closure union over-approximates the legal set; the firewall
  (`validate_tokens` / the commit path) rejects illegal tokens before they reach
  the sequence, so over-allowing is safe (the sampling distribution is
  preserved). For deterministic grammars the PDA mask is exact (the settled gate).
- **The terminal bridge**: the `terminal_token_map` is exact for the specific
  terminals (the special tokens, the single-byte literals, the positive ranges)
  and uses the DFA for the complex regex terminals. The PDA validate path is
  gated on `pda_bridge_exact` (the differential proof); when inexact it falls
  back to the speculative Earley walk.

## Benchmarks

Legacy `compute_mask` (the O(item_set) Earley path the PDA replaces), median ms
per call, from the `compute_mask` criterion bench (the cost grows with depth):

    depth 8192:    1.3353 ms
    depth 32768:   1.5021 ms
    depth 65536:   1.8196 ms
    depth 128000:  2.3494 ms

The PDA mask is O(kappa) (constant; ~107 states for the full-envelope
tool-calling grammar), so it does not grow with depth. At shallow depth the two
are comparable (the `pda_vs_legacy_mask_benchmark`, the 10-190 token range):

    depth | legacy (us/token) | PDA (us/token) | speedup
    10    | 1250.55           | 1244.85        | 1.0x
    50    | 1207.25           | 1161.70        | 1.0x
    100   | 1189.55           | 1146.70        | 1.1x
    190   | 1271.65           | 1184.20        | 1.1x

The advantage is depth-dependent: it is ~1x at short depth (the legacy
item-set is still small) and grows as the legacy set approaches the 50000-item
limit on long recursive sequences - needed for full coverage of long-reasoning
models and large `max_tokens` limits on outputs.

## Effects on single-token sampling and drafting

**Single-token sampling** (the `pda_fused_sample` kernel, the attention-rs):
one launch does mask + sample + advance. The PDA mask (the VOB) is applied to
the logits (illegal tokens -> -inf) before the greedy / top-k / top-p sample, and
the sampled token advances the PDA in the same launch (the bounded stack is
updated in place). If no grammar is active (`ctrl` is null) the kernel runs
plain sampling with zero PDA overhead.

**Drafting** (the `pda_fused_project` kernel, the MTP / DFlash path): one
launch walks K draft tokens through the PDA and emits the K+1 projected masks
(the start + after each draft token), applied to the K draft logit rows in the
same pass. This replaces K sequential `mask_at_cfg` calls (each an
O(epsilon_closure) BFS) with a single batched projection, so the drafting
constraint is free on the GPU.

## CUDA kernel snippets (from `attention-rs`)

The CSR-closure mask (matches the CPU `mask_at_cfg`):
```cuda
    __device__ void scan_mask(const uint32_t* transitions,
        const uint32_t* ctrl_u32_offsets, const uint32_t* ctrl_counts,
        uint32_t num_inputs, uint32_t ctrl, uint32_t top,
        uint32_t* out_vob, uint32_t words_per_vob) {
        for (uint32_t w = 0; w < words_per_vob; w++) out_vob[w] = 0;
        const int MAXF = 64;
        uint32_t fq[MAXF], ft[MAXF];
        int head = 0, tail = 0;
        fq[tail] = ctrl; ft[tail] = top; tail++;
        while (head < tail) {
            uint32_t cq = fq[head], ctop = ft[head]; head++;
            size_t idx = ctrl_u32_offsets[cq];
            uint32_t count = ctrl_counts[cq];
            for (uint32_t i = 0; i < count; i++) {
                uint32_t a = transitions[idx + 1];
                uint32_t t = transitions[idx + 2];
                uint32_t next_q = transitions[idx + 3];
                uint32_t push_len = transitions[idx + 4];
                if (t == ctop) {
                    if (a < num_inputs) {
                        uint32_t word_idx = a / 32;
                        if (word_idx < words_per_vob)
                            out_vob[word_idx] |= (1u << (a % 32));
                    } else {
                        uint32_t new_top = (push_len > 0) ? transitions[idx + 5] : ctop;
                        bool dup = false;
                        for (int v = 0; v < tail; v++)
                            if (fq[v] == next_q && ft[v] == new_top) { dup = true; break; }
                        if (!dup && tail < MAXF) { fq[tail] = next_q; ft[tail] = new_top; tail++; }
                    }
                }
                idx += 5 + push_len;
            }
        }
    }
```

The fused sample (mask + sample + advance in one launch, one thread per
sequence; the anti-loop kick clears the `forbid` bit with a popcount > 1 safety
floor so it never dead-ends the sequence):
```cuda
    template <typename T>
    __global__ void pda_fused_sample_kernel(const T* logits, const uint32_t* ctrl,
        uint32_t* stack, const uint32_t* sp, uint32_t* out_ctrl, uint32_t* out_sp,
        uint32_t* out_tokens, const uint32_t* transitions, const uint32_t* accepting,
        const uint32_t* ctrl_u32_offsets, const uint32_t* ctrl_counts,
        uint32_t* vob_buf, const uint32_t* forbid, uint32_t num_states,
        uint32_t num_stack_syms, uint32_t num_inputs, uint32_t words_per_vob,
        int batch, int vocab, int top_k, float temperature, float top_p,
        uint64_t seed, int d) {
        int seq = blockIdx.x * blockDim.x + threadIdx.x;
        if (seq >= batch) return;
        const T* logit_row = logits + (size_t)seq * vocab;
        uint32_t* vob = vob_buf + (size_t)seq * words_per_vob;
        uint32_t c = 0, s = 1, top = 0;
        if (ctrl != nullptr) {
            c = ctrl[seq];
            if (c >= num_states) c = 0;
            s = (sp != nullptr) ? sp[seq] : 1;
            top = (s > 0 && stack != nullptr) ? stack[(size_t)seq * d + s - 1] : 0;
            scan_mask(transitions, ctrl_u32_offsets, ctrl_counts, num_inputs, c, top, vob, words_per_vob);
            if (forbid != nullptr) {
                uint32_t f = forbid[seq];
                if (f != UINT32_MAX && f < num_inputs) {
                    uint32_t popcount = 0;
                    for (uint32_t w = 0; w < words_per_vob; w++) popcount += __popc(vob[w]);
                    if (popcount > 1) {
                        uint32_t word_idx = f / 32;
                        if (word_idx < words_per_vob) vob[word_idx] &= ~(1u << (f % 32));
                    }
                }
            }
        } else {
            for (uint32_t w = 0; w < words_per_vob; w++) vob[w] = 0xFFFFFFFF;
        }
        // ... greedy (top_k<=1, temperature<=0) or top-k + curand sample over the
        //     masked logits (illegal / below-kth -> -inf); then advance_pda writes
        //     out_ctrl / out_sp and pushes the symbols into the bounded stack.
    }
```

The drafting projection (K+1 masks in one launch; breaks on divergence, matching
the CPU `project_batch`):
```cuda
    __global__ void pda_fused_project_kernel(const uint32_t* ctrl, const uint32_t* stack,
        const uint32_t* sp, const uint32_t* drafts, uint32_t* out_masks,
        const uint32_t* transitions, const uint32_t* accepting,
        const uint32_t* ctrl_u32_offsets, const uint32_t* ctrl_counts,
        const uint32_t* forbid, uint32_t num_states, uint32_t num_stack_syms,
        uint32_t num_inputs, uint32_t words_per_vob, int batch, int k, int d) {
        int seq = blockIdx.x * blockDim.x + threadIdx.x;
        if (seq >= batch) return;
        uint32_t c = ctrl[seq];
        if (c >= num_states) c = 0;
        uint32_t s = (sp != nullptr) ? sp[seq] : 1;
        uint32_t top = (s > 0 && stack != nullptr) ? stack[(size_t)seq * d + s - 1] : 0;
        const uint32_t* draft_row = drafts + (size_t)seq * k;
        for (int pos = 0; pos <= k; pos++) {
            uint32_t* out = out_masks + ((size_t)seq * (k + 1) + pos) * words_per_vob;
            scan_mask(transitions, ctrl_u32_offsets, ctrl_counts, num_inputs, c, top, out, words_per_vob);
            if (pos == k) break;
            uint32_t tok = draft_row[pos];
            uint32_t next_c, next_top, next_push_len, moved;
            advance_pda(transitions, ctrl_u32_offsets, ctrl_counts, num_inputs,
                c, top, tok, &next_c, &next_top, &next_push_len, nullptr, &moved);
            if (!moved) break;   // the draft diverged (no transition)
            c = next_c; top = next_top;
        }
    }
```

## Test coverage

- `test_dpda` (15 proofs), `test_pda_differential` (3), `test_pda_sequence_growth`
  (4, including the PDA-vs-legacy benchmark), `test_golden_mask` (2) all green
  with the `dpda` feature.
- The full llguidance suite (nopda + pda/simdpda): 1553 + 1577 tests, 0 failed.
- The pushdown-rs suite (`--all-features`): green (the scalar/SIMD/GPU identity,
  the bitvec round-trip, the differential oracle).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -187,6 +187,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1b8e56985ec62d17e9c1001dc89c88ecd7dc08e47eba5ec7c29c7b5eeecde967"
 
 [[package]]
+name = "bitvec"
+version = "1.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ddcec3d12c579d40898fe0a9a358a803c23e9c52ca3c425707f81c9436211837"
+dependencies = [
+ "funty",
+ "radium",
+ "tap",
+ "wyz",
+]
+
+[[package]]
 name = "borrow-or-share"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -747,6 +759,12 @@ dependencies = [
  "lazy_static",
  "num",
 ]
+
+[[package]]
+name = "funty"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e6d5a32815ae3f33302d95fdcb2ce17862f8c65363dcfd29360480ba1001fc9c"
 
 [[package]]
 name = "futures"
@@ -1405,6 +1423,7 @@ dependencies = [
  "jsonschema",
  "lazy_static",
  "llg_test_utils",
+ "pushdown-rs",
  "rand",
  "rayon",
  "referencing",
@@ -1862,6 +1881,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "pushdown-rs"
+version = "0.1.0"
+source = "git+https://github.com/sempervictus/pushdown-rs#da365889af7522aa3573264a437228e2222a6b2c"
+dependencies = [
+ "bitvec",
+ "rten-simd",
+]
+
+[[package]]
 name = "pyo3"
 version = "0.29.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1933,6 +1961,12 @@ name = "r-efi"
 version = "5.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "69cdb34c158ceb288df11e18b4bd39de994f6657d83847bdffdbd7f346754b0f"
+
+[[package]]
+name = "radium"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dc33ff2d4973d518d823d61aa239014831e521c75da58e3df4840d3f47749d09"
 
 [[package]]
 name = "rand"
@@ -2170,6 +2204,12 @@ dependencies = [
  "syn",
  "unicode-ident",
 ]
+
+[[package]]
+name = "rten-simd"
+version = "0.26.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7f0c01294eb782adca256aa130b48a04006d03763073a68f45a9c33b982ea556"
 
 [[package]]
 name = "rustc-demangle"
@@ -2543,6 +2583,12 @@ dependencies = [
  "core-foundation-sys",
  "libc",
 ]
+
+[[package]]
+name = "tap"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "55937e1799185b12863d447f42597ed69d9928686b8d88a1df17376a097d8369"
 
 [[package]]
 name = "target-lexicon"
@@ -3413,6 +3459,15 @@ name = "writeable"
 version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ea2f10b9bb0928dfb1b42b65e1f9e36f7f54dbdf08457afefb38afcdec4fa2bb"
+
+[[package]]
+name = "wyz"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "05f360fc0b24296329c78fda852a1e9ae82de9cf7b27dae4b7f62f118f77b9ed"
+dependencies = [
+ "tap",
+]
 
 [[package]]
 name = "yoke"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1435,6 +1435,7 @@ dependencies = [
  "serde_json",
  "toktrie",
  "toktrie_hf_downloader",
+ "toktrie_hf_tokenizers",
 ]
 
 [[package]]
@@ -1883,7 +1884,7 @@ dependencies = [
 [[package]]
 name = "pushdown-rs"
 version = "0.1.0"
-source = "git+https://github.com/sempervictus/pushdown-rs#ceb709c24a93322b2aaa0ad1cc9499ccb4c082ff"
+source = "git+https://github.com/sempervictus/pushdown-rs#7cd3f7db05fa6286c2974dbba52b3d86c5e467f6"
 dependencies = [
  "bitvec",
  "rten-simd",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1883,7 +1883,7 @@ dependencies = [
 [[package]]
 name = "pushdown-rs"
 version = "0.1.0"
-source = "git+https://github.com/sempervictus/pushdown-rs#da365889af7522aa3573264a437228e2222a6b2c"
+source = "git+https://github.com/sempervictus/pushdown-rs#ceb709c24a93322b2aaa0ad1cc9499ccb4c082ff"
 dependencies = [
  "bitvec",
  "rten-simd",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1884,7 +1884,7 @@ dependencies = [
 [[package]]
 name = "pushdown-rs"
 version = "0.1.0"
-source = "git+https://github.com/sempervictus/pushdown-rs#7cd3f7db05fa6286c2974dbba52b3d86c5e467f6"
+source = "git+https://github.com/sempervictus/pushdown-rs?rev=461fb26#461fb2625ddbfe69421c8868a49721be3c479e8b"
 dependencies = [
  "bitvec",
  "rten-simd",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1884,7 +1884,7 @@ dependencies = [
 [[package]]
 name = "pushdown-rs"
 version = "0.1.0"
-source = "git+https://github.com/sempervictus/pushdown-rs?rev=461fb26#461fb2625ddbfe69421c8868a49721be3c479e8b"
+source = "git+https://github.com/sempervictus/pushdown-rs?rev=1361623#1361623a3f3e64e34231e78e247495dd5567094c"
 dependencies = [
  "bitvec",
  "rten-simd",
--- a/README.md
+++ b/README.md
@@ -81,6 +81,23 @@ The library implements a context-free grammar parser using Earley’s algorithm 
 
 Grammars can be also used to speed up decode via [fast-forward tokens](./docs/fast_forward.md).
 
+### GPU Offload (PDA)
+
+For inference runtimes with GPU sampling (e.g. xinfer, vLLM, TensorRT-LLM),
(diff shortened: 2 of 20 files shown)


hudson-ai:
Hi @sempervictus this is a very very cool idea! I sadly don't have a ton of bandwidth to review this at the moment though. Do you feel confident in its maturity at this point that you'd like some eyes on it? I saw that you were going back and forth across a few different approaches over the last week or two.

sempervictus:
Sorry for lag @hudson-ai - conference travel. I'd put this at "if people more familiar with parser/lexer/dfa/fsm logic have a few cycles to look at it, will probably help it mature" level of interest. I'll be doing more testing this week and hopefully finding some edge cases and optimizations but i think the idea of "geometrically projecting where the FSM can go" has legs both for performance purposes in conventional single-token-per-pass sampling and both linear & parallel drafting attempts.

PR #385, opened 2026-09-21
Move ffi implementation to a different crate

madsmtm:
This:
- Avoids redundant creation of dylib and staticlib. I hit this when compiling for Emscripten, where I was trying to set a linker flag for the root crate, and that flag ended up being set when building `llguidance.so` (and then failed).
- Avoids the `#[no_mangle]` symbols. These make it impossible to have two different versions of `llguidance` in the same binary, which in turn probably makes releasing a `llguidance v2.0.0` infeasible (or at least quite annoying for the ecosystem).
  - This also makes it possible for the linker to dead-code eliminate `llguidance` (`#[no_mangle]` forces the linker to keep the symbols in the final binary).

While still allowing using the library as a whole from C (though users will have to change their compilation setup slightly to point to `c/llguidance.h` instead of `parser/llguidance.h`).


madsmtm:
Let me know if that description was unclear!

I'm fairly convinced that the general approach (don't set `crate-type = ...` in the main `llguidance` crate) is the right approach, but I'm also up for changing to whatever suits your needs?

hudson-ai:
Hey @madsmtm, thanks for the contribution! I agree with the goal of separating the C ABI from the Rust crate :)

My concern is that this removes the public `llguidance::ffi` API and we don't currently have plans for a `2.0` release. This public API is used by [llguidance_bg](https://github.com/guidance-ai/llguidance_bg/tree/main), so it's a concrete compatibility break.

Is there a way to preserve the existing Rust-facing FFI API while moving the C symbol exports and library artifacts into the new `llguidance_c` crate? I think that the new crate would also need a supported distribution path for C consumers -- currently it's marked `publish = false`

madsmtm:
Ah, I hadn't noticed that it's public API.

I think the best way to support that would probably be to instead change this PR to:
1. Add a new non-default feature flag `ffi-no-mangle` and cfg-gate things with `#[cfg_attr(feature = "ffi-no-mangle", no_mangle)]`.
2. Add `llguidance_c` containing just `crate-type = ["dylib", "staticlib"]` and `llguidance = { .., features = ["ffi-no-mangle"] }`.

Though if `llguidance_bg` is relying on the `#[no_mangle]` (it's not immediately clear to me whether they are?), then we'd need to make the feature flag be inverted (`no-ffi-no-mangle`) (and ideally add it to the default features if we can stomach that tiny break, that'd work for `llguidance_bg` which uses `default-features = false`, and would be the nicer experience for everyone else).

WDYT?

> I think that the new crate would also need a supported distribution path for C consumers -- currently it's marked `publish = false`

What is your distribution path like right now, apart from building from source? `cargo install llguidance` doesn't work.

